PR #62, opened 2022-12-04
Class22 83-confirmation-message-for-valid-sign-up-form

Hailegebriel22:
# Description

Message for a valid sign-up form
Fixes # (issue)

# How to test?

Please provide a short summary how your changes can be tested?

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have followed the name conventions for CSS Classnames and filenames, Components names and filenames, Style filenames, if you are in doubt check the the project README.MD and here https://github.com/HackYourFuture-CPH/curriculum/blob/master/review/review-checklist.md
- [ ] I have commented my code, particularly in hard-to-understand areas, if you code was simple enough mark the box anyway
- [ ] I have made corresponding changes to the documentation, if you code was simple enough mark the box anyway
- [ ] This PR is ready to be merged and not breaking any other functionality


Hanouj (dismissed):

packages/client/src/containers/Login.container/SignupForm.js
```diff
@@ -9,14 +9,25 @@ const SignupForm = () => {
...
   const [formErrors, setFormErrors] = useState({});
+  const [isSubmit, setIsSubmit] = useState(false);
+  const message = '';
+
   const handleChange = (e) => {
     setFormValues({ ...formValues, [e.target.name]: e.target.value });
   };
 
   const handleSubmit = (event) => {
+    setIsSubmit(true);
```
annamourasobral, line 20: submit can only be allowed **after** a conditional validation being passed for best practice

packages/client/src/containers/Login.container/SignupForm.js
```diff
@@ -9,14 +9,25 @@ const SignupForm = () => {
     password: '',
   });
   const [formErrors, setFormErrors] = useState({});
+  const [isSubmit, setIsSubmit] = useState(false);
+  const message = '';
```
annamourasobral, line 13: this is not necessary

packages/client/src/containers/Login.container/SignupForm.js
```diff
@@ -52,14 +62,18 @@ const SignupForm = () => {
...
         className="signup-submit-button"
         type="submit"
       >
         Sign up
-      </button>
+      </button>{' '}
+      {Object.keys(formErrors).length === 0 && isSubmit ? (
+        <p className="success-message">Thank you for Signing up</p>
+      ) : (
+        <p>{message}</p>
```
annamourasobral, line 75:
```suggestion
        <></>
```

packages/client/src/containers/Login.container/SignupForm.js
```diff
@@ -9,14 +9,25 @@ const SignupForm = () => {
...
   const handleSubmit = (event) => {
+    setIsSubmit(true);
     event.preventDefault();
     setFormErrors(FormValidation(formValues));
   };
+  /* eslint-disable no-console */
+
+  useEffect(() => {
+    if (Object.keys(formErrors).length === 0 && isSubmit) {
+      console.log(formValues);
```
annamourasobral, line 28: do you really need to keep tracking of formValues on the console?

packages/client/src/containers/Login.container/SignupForm.js
```diff
@@ -52,14 +62,18 @@ const SignupForm = () => {
...
       )}
-
       <button
         onClick={handleSubmit}
         className="signup-submit-button"
         type="submit"
       >
         Sign up
-      </button>
+      </button>{' '}
```
annamourasobral, line 71:
```suggestion
      </button>
```

packages/client/src/containers/Login.container/SignupForm.js
```diff
@@ -52,14 +62,18 @@ const SignupForm = () => {
...
+      {Object.keys(formErrors).length === 0 && isSubmit ? (
```
annamourasobral, line 72: this will be redundant once you refactor the handler
```suggestion
      { isSubmit ? (
```

annamourasobral (changes requested):
Hey @Hailegebriel22 
I suggested a refactoring so you only allow the user to go forward if there are no errors in the form and the success message was rendered. 
Keep going!

packages/client/src/containers/Login.container/SignupForm.js
Hailegebriel22: I get the error " there‘s no need for a Fragment at all" whenever I use <></> 
that is why I add an empty string.

packages/client/src/containers/Login.container/SignupForm.js
Hailegebriel22: One of the tasks is to log form content into the console. or should I do it differently?

packages/client/src/containers/Login.container/SignupForm.js
annamourasobral: you can also use ('')

packages/client/src/containers/Login.container/SignupForm.js
annamourasobral: Ok, sorry I cannot check Jira. 
I just wondered so we don't have an overload in the console.

Hanouj (approved):

packages/client/src/containers/Login.container/SignupForm.js
```diff
@@ -10,24 +10,16 @@ const SignupForm = () => {
   });
   const [formErrors, setFormErrors] = useState({});
   const [isSubmit, setIsSubmit] = useState(false);
-  const message = '';
 
   const handleChange = (e) => {
     setFormValues({ ...formValues, [e.target.name]: e.target.value });
   };
 
   const handleSubmit = (event) => {
```
annamourasobral, line 18: I meant that:
```
const handleSubmit = (event) => {
 event.preventDefault();
 if(Object.keys(formErrors).length === 0) {
 setIsSubmit(true);
 }
  }; 
```
Validations are not supposed to be made on submit time, but after the user types the input. Usually, we use the onBlur propertie to run a validation function and then reset it onFocus.

annamourasobral (changes requested):

packages/client/src/containers/Login.container/SignupForm.js
miltonhultgren: Or just `null` if you don't want to render anything at all.

packages/client/src/containers/Login.container/SignupForm.js
Hailegebriel22: @annamourasobral  okay Thanks, 
I need to add the form validation as well , right?
```
const handleSubmit = (event) => {
 event.preventDefault();
**setFormErrors(FormValidation(formValues));**
 if(Object.keys(formErrors).length === 0) {
 setIsSubmit(true);
 }
  }; 
```

annamourasobral (approved):

Hanouj (approved):

PR #63, opened 2022-12-04
feature(Functional Categories Dropdown)

patel-prakashkumar:
# Description
I have created Functional Categories Dropdown.
Please provide a short summary explaining what this PR is about.

Fixes # (issue)

# How to test?

Please provide a short summary how your changes can be tested?

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have followed the name conventions for CSS Classnames and filenames, Components names and filenames, Style filenames, if you are in doubt check the the project README.MD and here https://github.com/HackYourFuture-CPH/curriculum/blob/master/review/review-checklist.md
- [ ] I have commented my code, particularly in hard-to-understand areas, if you code was simple enough mark the box anyway
- [ ] I have made corresponding changes to the documentation, if you code was simple enough mark the box anyway
- [ ] This PR is ready to be merged and not breaking any other functionality


packages/client/src/components/ProductListDropDown/DropDownViewList/DropDownViewList.component.js
```diff
@@ -1,39 +1,59 @@
...
-    'Reviews',
-    'Brand',
-  ];
+const DropDownViewList = ({
+  selectCategory,
+  selectSortBy,
+  selectAllFilter,
+}) => {
+  const [categories, setCategories] = useState([]);
+  const fetchProducts = () => {
```
miltonhultgren, line 13: Probably this should be named fetchCategories, but beyond that, why is this component fetching categories? Would it not make more sense if the container did that job and gave this component the data it needed to render?

packages/client/src/components/ProductListDropDown/DropDownViewList/DropDownViewList.component.js
```diff
@@ -1,39 +1,59 @@
...
+      .then((res) => res.json())
+      .then((item) => {
+        setCategories(item);
+      });
+  };
+  useEffect(() => {
+    fetchProducts();
+  }, []);
+
+  const Categories = categories.map(
```
miltonhultgren, line 24: Name your variables in camelCase, not PascalCase :) So just `categories`.

packages/client/src/components/ProductListDropDown/DropDownViewList/DropDownViewList.component.js
```diff
@@ -1,39 +1,59 @@
...
+        options={AllFilter}
+        select={selectAllFilter}
+      />
     </div>
   );
 };
+DropDownViewList.propTypes = {
+  selectCategory: PropTypes.func.isRequired,
+  selectSortBy: PropTypes.func.isRequired,
+  selectAllFilter: PropTypes.func.isRequired,
```
miltonhultgren, line 56: These are event handlers, so they are called **when** something happens but they don't decide **what** happens. 

I would advice naming them `onCategorySelected` instead, so that when this component is used it looks like this:
`<DropDownViewList onCategorySelected={setCategory} />`

packages/client/src/components/ProductList/ProductList.component.js
```diff
@@ -22,6 +23,37 @@ export default function ProductList() {
...
+  const fetchDropdown = (categoryId) => {
+    fetch(`${apiURL()}/products?category=${categoryId}`)
+      .then((res) => res.json())
+      .then((item) => {
+        setProducts(item);
+      });
+  };
+  useEffect(() => {
+    fetchDropdown(category);
+  }, [category]);
```
miltonhultgren, line 35: Again, this should probably be the responsibility of the container, same for the sorting. 

The component should inform the container of what should happen, the container knows how to do it.

miltonhultgren (changes requested):
I would like us to move around the data fetching responsibilities so that those are kept with the containers but overall the code itself looks fine :)

packages/client/src/components/ProductListDropDown/DropDownViewList/DropDownViewList.component.js
patel-prakashkumar: Thank you for your feedback. I have changed according to feedback.

packages/client/src/components/ProductListDropDown/DropDownViewList/DropDownViewList.component.js
patel-prakashkumar: I have updated name of variable.

packages/client/src/components/ProductListDropDown/DropDownViewList/DropDownViewList.component.js
patel-prakashkumar: I have change useState name according to feedback.

packages/client/src/components/ProductList/ProductList.component.js
patel-prakashkumar: I have moved fetch function and logic to collection page in container.

patel-prakashkumar: I have moved fetch Api function in container. just remain to move sort function from productList components to container. Its come error while moving to container because of search input functionality.

packages/client/src/components/CategoriesListDropDown/CategoriesListDropDown.component.js
```diff
@@ -1,25 +1,34 @@
...
 import PropTypes from 'prop-types';
-import './DropDownView.style.css';
+import './CategoriesListDropDown.style.css';
 
-const DropDownView = ({ options, lable, showFilterIcon, ...props }) => {
+const DropDownView = ({ options, lable, select, showFilterIcon, ...props }) => {
   const [value, setValue] = useState('');
   const handleChange = (event) => {
     setValue(event.target.value);
+    select(event.target.value);
```
miltonhultgren: I'd rather name this `onSelect`, since the component itself doesn't know what happens. It just knows to "call the callback on select"

packages/client/src/components/ProductList/ProductList.component.js
```diff
@@ -1,29 +1,31 @@
 import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
 import './ProductList.component.css';
 import ProductCard from '../ProductCard/ProductCard.component';
-import { apiURL } from '../../apiURL';
 import SearchInput from '../SearchInput/SearchInput.component';
 
-export default function ProductList() {
-  const [products, setProducts] = useState([]);
+export default function ProductList({ isLoading, products, onSelectedSortBy }) {
```
miltonhultgren: `onSelectedSortBy`, this is an actual value, not a callback. So I think here it makes more sense to call it `sortOrder`.

packages/client/src/components/ProductList/ProductList.component.js
```diff
@@ -1,29 +1,31 @@
...
+      }
+      if (sort === 'Price ↓') {
+        return Number(a.price) > Number(b.price) ? 1 : -1;
+      }
+      if (sort === 'Price ↑') {
+        return Number(a.price) < Number(b.price) ? 1 : -1;
+      }
+      return 0;
+    });
   };
```
miltonhultgren:
```
useEffect(() => {
  function sortProducts(a, b) {
    if (sortOrder === '') {
      return 0;
    }
    if (sortOrder === 'Alphabetically') {
      return a.name.toLowerCase() < b.name.toLowerCase() ? -1 : 1;
    }
    if (sortOrder === 'Price ↓') {
      return Number(a.price) > Number(b.price) ? 1 : -1;
    }
    if (sortOrder === 'Price ↑') {
      return Number(a.price) < Number(b.price) ? 1 : -1;
    }
    return 0;
  }

  filteredProducts.sort(sortFunction);
}, [sortOrder]);
```

I think you want to use an effect here instead to react to when the prop changes.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -1,12 +1,91 @@
...
-import DropDownViewList from '../../components/ProductListDropDown/DropDownViewList/DropDownViewList.component';
+import './CollectionsPage.style.css';
+import DropDownView from '../../components/CategoriesListDropDown/CategoriesListDropDown.component';
+import { apiURL } from '../../apiURL';
 
 export const CollectionsPage = () => {
+  const [products, setProducts] = useState([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [onSelectedCategory, setOnSelectedCategory] = useState('');
+  const [onSelectedSortBy, setOnSelectedSortBy] = useState('');
```
miltonhultgren: These are not event handlers, these are just values.
```suggestion
  const [selectedCategory, setSelectedCategory] = useState('');
  const [sortOrder, setSortOrder] = useState('');
```

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -1,12 +1,91 @@
...
+      .then((item) => {
+        setProducts(item);
+      })
+      .then(() => {
+        setIsLoading(false);
+      });
+  };
+  useEffect(() => {
+    fetchProducts();
+  }, []);
```
miltonhultgren:
```suggestion
  useEffect(() => {
    const fetchProducts = () => {
      setIsLoading(true);
      fetch(`${apiURL()}/products`)
        .then((res) => res.json())
        .then((item) => {
          setProducts(item);
          setIsLoading(false);
        });
    };
    
    fetchProducts();
  }, []);
```

Prefer to put the logic of the effect inside of the effect so it's clear that this happens as an effect.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -1,12 +1,91 @@
...
+
+  const [category, setCategoy] = useState([]);
```
miltonhultgren: Prefer to group your state at the top of your component so it's easy to see what data this component owns.
Also, this should probably be `categories` since it will be many?

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -1,12 +1,91 @@
...
+          // select={(allFilterName) => setOnSelectedAllFilter(allFilterName)}
+          showFilterIcon
+        />
+      </div>
+
+      <div>
+        <ProductList
+          isLoading={isLoading}
+          products={products}
+          onSelectedSortBy={onSelectedSortBy}
```
miltonhultgren: Still feel like the sorting should happen in this component since the stats (products and sort order) live here already.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -1,12 +1,91 @@
...
+  };
+  useEffect(() => {
+    fetchCategories();
+  }, []);
+
+  const categories = category.map(
+    (element) =>
+      element.name.charAt(0).toUpperCase() +
+      element.name.slice(1).toLowerCase(),
+  );
```
miltonhultgren:
```suggestion
const [categories, setCategories] = useState([]);

useEffect(() => {
  const fetchCategories = () => {
    fetch(`${apiURL()}/categories`)
      .then((res) => res.json())
      .then((items) => {
        setCategories(items.map(
          (item) =>
            item.name.charAt(0).toUpperCase() +
            item.name.slice(1).toLowerCase(),
        ));
      });
  };

  fetchCategories();
}, []);
```

You'll have a problem here though that `categories` will be a list of strings, how will you get access to the ID?

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -1,12 +1,91 @@
...
+
+  const sortBy = ['Recent Collections', 'Alphabetically', 'Price ↓', 'Price ↑'];
+  const allFilter = ['Price', 'Size', 'Color', 'Reviews', 'Brand'];
```
miltonhultgren: Maybe `sortOptions` and `filterOptions`?

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -1,12 +1,91 @@
...
+  const fetchDropdown = (categoryId) => {
+    fetch(`${apiURL()}/products?category=${categoryId}`)
+      .then((res) => res.json())
+      .then((item) => {
+        setProducts(item);
+      });
+  };
+  useEffect(() => {
+    fetchDropdown(onSelectedCategory);
+  }, [onSelectedCategory]);
```
miltonhultgren: This part still confuses me, we fetch products twice in different places.

I would instead merge this into one place but check if you can add the category filter if it's set, something like:
```
useEffect(() => {
  function fetchProducts() {
    // Still, how do you get the category ID here?
    const url = `${apiURL()}/products${selectedCategory ? '?category=${categoryId}' : ''}`;
  
    fetch(url)
      .then((res) => res.json())
      .then((items) => {
        setProducts(items);
      });
  };

  fetchProducts();
}, [selectedCategory]);
```

patel-prakashkumar: Its my mistake. Its fetch category Name not categoryId  but Yea, I will use useEffect for render product by category name. I will try to add category filter into one fetch products.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
patel-prakashkumar: I having error if I used sorting function in collection page because of serach input.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
patel-prakashkumar: I have merge product and query param in one fetch.  I have passed category name as fetch function parameter. 
useEffect(() => {
    function fetchProducts(categoryName) {
      const url = `${apiURL()}/products${selectedCategory ? `?category=${categoryName}` : ''}`;
      fetch(url)
        .then((res) => res.json())
        .then((items) => {
          setProducts(items);
        });
    }
    fetchProducts(selectedCategory);
  }, [selectedCategory]);

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
miltonhultgren: Then let's look at that error and try to fix that instead. Can you share the error?

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
miltonhultgren: That's better, though you still have the problem that the API expects the ID, not the name, right?

miltonhultgren (approved):
Approving for now, let's create issues to fix the category filter bug (need the ID) and the sorting logic bug. 

Thanks for fixing so much though!

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
patel-prakashkumar: Yes.. that true.. but we create endpoint by name.

PR #70, opened 2022-12-05
CLASS22-84-restyling-product-view-page-with-adding-reviews

RZhyvitskyi:
# Description:
fixes esponsiveness of Product View page and adds reviews 

# Jira:


viedamayanti (approved):

miltonhultgren (dismissed):
I think you've gone too far in how atomic you're making your components.

Don't break things down too far. If a component doesn't make sense on it's own it shouldn't be a component. So the back button for a carousel, does that make sense alone? Alone it's just a button and we already have a component for that. So I would say it doesn't. 

Same for many of the wrapper components that are added to just hold a CSS class. Figure out ways to add that CSS class to the wrapped component instead.

In general I would say if you need to use * imports you have too many components!

The code in each component itself is fine but I think it would be much easier to understand if it was more collected. 
Remember, code is read more often than it is written!

viedamayanti (approved):

RZhyvitskyi:
> I think you've gone too far in how atomic you're making your components.
> 
> Don't break things down too far. If a component doesn't make sense on it's own it shouldn't be a component. So the back button for a carousel, does that make sense alone? Alone it's just a button and we already have a component for that. So I would say it doesn't.
> 
> Same for many of the wrapper components that are added to just hold a CSS class. Figure out ways to add that CSS class to the wrapped component instead.
> 
> In general I would say if you need to use * imports you have too many components!
> 
> The code in each component itself is fine but I think it would be much easier to understand if it was more collected. Remember, code is read more often than it is written!

RZhyvitskyi:
I'll make all the changes regarding to Milton review soon, just waiting for Carousel to be done.

viedamayanti (approved):

ShravyaGurmitkal (approved):

shpomp:
Again, @RZhyvitskyi as per common guidelines, you need a peer and a mentor review, which != 2 peers. 👌🏽

PR #71, opened 2022-12-05
feature(migration): added migration of orders and order_items table

27ManishaPatel:
# Description:
added migration of orders and order_item table

# Jira:

### DIFF
--- a/packages/server/migrations/20221205220007_add_order_items_table.js
+++ b/packages/server/migrations/20221205220007_add_order_items_table.js
@@ -4,7 +4,6 @@
  */
 exports.up = function (knex) {
   return knex.schema.createTable('order_items', (table) => {
-    table.increments();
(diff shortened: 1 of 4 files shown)


packages/server/migrations/20221205220007_add_order_items_table.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.createTable('order_items', (table) => {
+    table.increments();
```
shpomp, line 7: this is not needed ✔️

shpomp (changes requested):

Yuliia-Balandiuk (approved):

shpomp (approved):

PR #72, opened 2022-12-06
style(DropDownView): add svg icon to the select

Yuliia-Balandiuk:
# Description:

# Jira:


packages/client/src/components/ProductListDropDown/DropDownView/DropDownView.style.css
```diff
@@ -9,9 +9,16 @@
   padding: 18px 12px 18px 12px;
   font-size: 18px;
   font-weight: 650;
+  position: relative;
   color: #212121cc;
-  background-color: #e4c1b1;
 }
+
+.all-filters {
+  background: url('../../../assets/all-flter-icon.svg') no-repeat;
```
miltonhultgren, line 17: This path should be absolute from the web server. So just `all-flter-icon.svg`.

Remember, all files in `assets` are served from the web server, so you shouldn't try to read it from the local git repo.

miltonhultgren (changes requested):
What does "extra" mean in this context? 
Extra options? Extra cheese? Extra CSS?

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -5,7 +5,7 @@ import DropDownViewList from '../../components/ProductListDropDown/DropDownViewL
 export const CollectionsPage = () => {
   return (
     <div>
-      <DropDownViewList extra />
+      <DropDownViewList filterIcon />
```
miltonhultgren: This is a bit better but maybe this should be something like "showFilterIcon" since it's a boolean after all. 
But I also wonder why it's hard coded here to true?

Also ask yourself if it really makes sense to name this prop the same in all the layers it's being passed through, does it mean the same basic thing in all those components? It's okay to rename them so that they make the most sense in the context of each individual component.

packages/client/src/components/ProductListDropDown/DropDownView/DropDownView.style.css
```diff
@@ -14,7 +14,7 @@
 }
 
 .all-filters {
-  background: url('../../../assets/all-flter-icon.svg') no-repeat;
+  background: url('all-filter-icon.svg') no-repeat;
```
miltonhultgren: I looked into this and this has changed since the last FP I mentored, the way you did it at first is indeed the correct way so sorry for the noise, please revert this line back to:
`background: url('../../../assets/all-filter-icon.svg') no-repeat;`

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
Yuliia-Balandiuk: Thanks for the help with the name for prop.
In my opinion, the same name, on the contrary, makes it easier to understand what kind of prop it is. It is always true because it is always used for one select.
Perhaps I did not understand something from your message, so I would be grateful for an explanation

packages/client/src/components/ProductListDropDown/DropDownView/DropDownView.component.js
```diff
@@ -15,9 +15,10 @@ const DropDownView = ({ options, lable, ...props }) => {
 
   return (
     <select
-      className="view-dropdown-select"
       value={value}
+      showFilterIcon={showFilterIcon}
```
miltonhultgren, line 19: Nitpick: This isn't a valid prop for a `select`

miltonhultgren (approved):

27ManishaPatel (approved):

PR #77, opened 2022-12-07
Class22 90 favorites migration and seeds

Hanouj:
# Description:

# Jira:


shpomp:
@Hanouj the build fail is because, in seeds, you use user IDs that do not exist. Look at user seeds to understand how many users we have. ✌🏽

PR #82, opened 2022-12-07
feature(Realted Items): made changes to handle the number of items in…

ShravyaGurmitkal:
… the Relatedites and moving carousel

# Description:

# Jira:

https://hackyourfuture-dk.atlassian.net/browse/CLASS22-63

### DIFF
--- a/packages/client/src/containers/RelatedItems/RelatedItems.Container.js
+++ b/packages/client/src/containers/RelatedItems/RelatedItems.Container.js
@@ -1,43 +1,73 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import ProductCard from '../../components/ProductCard/ProductCard.component';
 import { apiURL } from '../../apiURL';
+import { ViewAfterCarousel } from '../../components/ViewCarousel/ViewForwadCarousel/ViewAfterCarousel';
 import './RelatedItems.styles.css';
 import { ViewBackCarousel } from '../../components/ViewCarousel/ViewBackCarousel/ViewBackCarousel.component';
-import { ViewForwardCarousel } from '../../components/ViewCarousel/ViewForwardCarousel/ViewForwardCarousel.component';
 
 export const RelatedItems = ({ category }) => {
   const [categoryProducts, setCategoryProducts] = useState([]);
+  const [offSet, setOffset] = useState(0);
+  const [dataFlag, setDataFlag] = useState(true);
+
+  const fetchProductsByCategory = useCallback(async () => {
+    const response = await fetch(
+      `${apiURL()}/products?category=${category}&limit=3&offset=${offSet}`,
+    );
+    const categoryProductData = await response.json();
+
+    if (categoryProductData.data.length < 3) {
+      fetchAllProducts();
+    } else {
+      setDataFlag(true);
+      setCategoryProducts(categoryProductData.data);
+    }
+  }, [category, offSet]);
+
+  const fetchAllProducts = async () => {
+    const response = await fetch(`${apiURL()}/products`);
+    const allProductsData = await response.json();
+    const productsData = allProductsData.slice(0, 3);
+    setCategoryProducts(productsData);
+    setDataFlag(false);
+  };
 
   useEffect(() => {
-    const fetchProductsByCategory = async () => {
-      const response = await fetch(`${apiURL()}/products?category=${category}`);
-      const categoryProductData = await response.json();
-
-      setCategoryProducts(categoryProductData);
-    };
-
     fetchProductsByCategory();
-  }, [category]);
+  }, [category, fetchProductsByCategory]);
+
+  const handleBackCarousel = () => {
+    if (offSet > 2) {
+      setOffset(offSet - 3);
+    }
+  };
+
+  const handleAfterCarousel = () => {
+    setOffset(offSet + 3);
+  };
 
   return (
     // TODO: will refactor function Number of items to display below. CLASS22-63"
     <div className="related-items">
       <h1 className="relatedItems-heading">Related Items</h1>
       <div className="related-items-display">
-        <div className="related-items-back">
-          <ViewBackCarousel />
+        <div className="back-carousel">
+          <ViewBackCarousel text="" onClick={handleBackCarousel} />
         </div>
         {categoryProducts.map((prod) => (
           <ProductCard
             key={prod.id}
+            product={prod}
+            rating={5}
             title={prod.name}
             price={prod.price}
-            rating={5}
           />
         ))}
-        <div className="related-items-forward">
-          <ViewForwardCarousel />
+        <div className="after-Carousel">
+          {dataFlag && (
+            <ViewAfterCarousel text="" onClick={handleAfterCarousel} />
+          )}
         </div>
       </div>
     </div>
--- a/packages/client/src/containers/RelatedItems/RelatedItems.Container.js
+++ b/packages/client/src/containers/RelatedItems/RelatedItems.Container.js
@@ -11,31 +11,32 @@ export const RelatedItems = ({ category }) => {
   const [offSet, setOffset] = useState(0);
   const [hasMoreProducts, setHasMoreProducts] = useState(true);
 
-  const fetchProductsByCategory = useCallback(async () => {
-    const response = await fetch(
-      `${apiURL()}/products?category=${category}&limit=3&offset=${offSet}`,
-    );
-    const categoryProductData = await response.json();
-
-    if (categoryProductData.data.length < 3) {
-      fetchAllProducts();
-    } else {
-      setHasMoreProducts(true);
-      setCategoryProducts(categoryProductData.data);
-    }
-  }, [category, offSet]);
-
-  const fetchAllProducts = async () => {
-    const response = await fetch(`${apiURL()}/products`);
-    const allProductsData = await response.json();
-    const productsData = allProductsData.slice(0, 3);
-    setCategoryProducts(productsData);
-    setHasMoreProducts(false);
-  };
-
   useEffect(() => {
+    async function fetchProductsByCategory() {
+      const response = await fetch(
+        `${apiURL()}/products?category=${category}&limit=3&offset=${offSet}`,
+      );
+      const categoryProductData = await response.json();
+      if (categoryProductData.length === 0) {
+        if (categoryProductData.length === 0 && offSet === 0) {
+          await fetchAllProducts();
+        }
+        setHasMoreProducts(false);
+      } else if (categoryProductData.length < 3) {
+        setHasMoreProducts(false);
+        setCategoryProducts(categoryProductData);
+      } else {
+        setHasMoreProducts(true);
+        setCategoryProducts(categoryProductData);
+      }
+    }
+    async function fetchAllProducts() {
+      const response = await fetch(`${apiURL()}/products`);
+      const products = await response.json();
+      setCategoryProducts(products.slice(0, 3));
+    }
     fetchProductsByCategory();
-  }, [category, fetchProductsByCategory]);
+  }, [category, offSet]);
 
   const handleBackCarousel = () => {
     if (offSet > 2) {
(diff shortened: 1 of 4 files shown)


RZhyvitskyi (dismissed):

miltonhultgren:
Here is a good article on how to write better commit messages https://cbea.ms/git-commit/

As an example, consider:
`feat(RelatedItems): Handle any number of items, Move carousel`

packages/server/api/routes/products.router.js
```diff
@@ -37,25 +37,29 @@ const productsController = require('../controllers/products.controller');
...
-  }
-  try {
-    productsController
-      .getProducts()
-      .then((result) => res.json(result))
-      .catch(next);
-  } catch (error) {
-    res.status(404).json({ error: 'Bad Get Request' });
+  } else {
+    try {
```
miltonhultgren, line 55: I'd wrap the whole router in this try/catch, any of those functions could throw.

Another thing to doubt about is why there are 3 different code paths here, what if I want to get things in category "shoes" but also search for "skateboarding"? 
Will you add a new if/else branch to check if I have both category and search in my query? 
I think it would be better if you figure out ways to support multiple options within the same request.
That's possibly out of scope for this PR but something to discuss with your team!

packages/server/api/controllers/products.controller.js
```diff
@@ -40,16 +40,25 @@ const getProductsByCategory = async (category) => {
...
+      status: 200,
+      data: products,
+    };
   } catch (error) {
-    return error.message;
+    return {
+      status: error.httpStatus,
+      data: [],
+      message: error.message,
+    };
```
miltonhultgren: This controller now mixes error handling patterns, here we return an error response, but on line 47 we throw a HttpError, wouldn't it be better to use one style?

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
```diff
@@ -1,33 +1,74 @@
...
+    setOffset(offSet + 3);
+  };
 
   return (
     // TODO: will refactor function Number of items to display below. CLASS22-63"
     <div className="related-items">
       <h1 className="realtedItems-heading">Related Items</h1>
       <div className="related-items-display">
+        <div className="back-carousel">
+          <ViewBackCarousel text="" onClick={handleBackCarousel} />
```
miltonhultgren: If it's not possible to go around backwards, then it would be better to disable this button if you're at the start of the items. Same for the next, if there is no more items, better to disable the button.

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
```diff
@@ -1,33 +1,74 @@
...
 import ProductCard from '../../components/ProductCard/ProductCard.component';
 import { apiURL } from '../../apiURL';
+import { ViewBackCarousel } from '../../components/ViewCarousel/ViewBackCarousel/ViewBackCarousel';
+import { ViewAfterCarousel } from '../../components/ViewCarousel/ViewForwadCarousel/ViewAfterCarousel';
 import './RelatedItems.styles.css';
 
 export const RelatedItems = ({ category }) => {
   const [categoryProducts, setCategoryProducts] = useState([]);
+  const [offSet, setOffset] = useState(0);
+  const [dataFlag, setDataFlag] = useState(true);
```
miltonhultgren: This variable name does not tell me at all what this boolean does.

After carefully reading the code, I **think** it means "hasMoreProducts", meaning there are more products to load from the API for the next "pages". 

Try to write code that doesn't make the reader think and read so carefully, the more you can communicate upfront the better.

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
```diff
@@ -1,33 +1,74 @@
...
+    const response = await fetch(
+      `${apiURL()}/products?category=${category}&limit=3&offset=${offSet}`,
+    );
+    const categoryProductData = await response.json();
+
+    if (categoryProductData.data.length < 3) {
+      fetchAllProducts();
+    } else {
+      setDataFlag(true);
+      setCategoryProducts(categoryProductData.data);
```
miltonhultgren: What if there is exactly 3 products? It will set the flag to true, but when you click next you'll get nothing back. 

Instead, it would be great if your API actually returned a count as well. So the response could look like:
```
{
  status: 200,
  total: 13,
  data: [
    product1,
    product2,
    product3
  ].
}
```

That way you don't need the flag, you can simply check if `total` is larger than `offset` .

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
```diff
@@ -1,33 +1,74 @@
...
-      const response = await fetch(`${apiURL()}/products?category=${category}`);
-      const categoryProductData = await response.json();
+  const fetchProductsByCategory = useCallback(async () => {
+    const response = await fetch(
+      `${apiURL()}/products?category=${category}&limit=3&offset=${offSet}`,
+    );
+    const categoryProductData = await response.json();
+
+    if (categoryProductData.data.length < 3) {
+      fetchAllProducts();
```
miltonhultgren: So we don't show the 2 related items we did find? Perhaps it would be better to show those and then say there isn't any more related items, or if it's 0 say there is no related items but show some other products?

miltonhultgren: You also probably want to use `await` here.

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
```diff
@@ -1,33 +1,74 @@
...
-    };
+  const fetchAllProducts = async () => {
+    const response = await fetch(`${apiURL()}/products`);
+    const allProductsData = await response.json();
+    const productsData = allProductsData.slice(0, 3);
+    setCategoryProducts(productsData);
+    setDataFlag(false);
+  };
 
+  useEffect(() => {
```
miltonhultgren: I would advice to wrap the logic for the effect inside of the effect, this makes it more clear where these functions belong.

```suggestion
useEffect(() => {
  async function fetchProductsByCategory() {
    const response = await fetch(
      `${apiURL()}/products?category=${category}&limit=3&offset=${offSet}`,
    );
    const { data: products } = await response.json();

    if (products.length < 3) {
      await fetchAllProducts();
    } else {
      setDataFlag(true);
      setCategoryProducts(products);
    }
  }

  async function fetchAllProducts() {
    const response = await fetch(`${apiURL()}/products`);
    const products = await response.json();
    setCategoryProducts(products.slice(0, 3));
    setDataFlag(false);
  }

  fetchProductsByCategory();
}, [category, offset]);
```

Where you might want to use `useCallback` is on the two click handlers below.

Another thing that makes me wonder, why does `api/products` return a different format than `api/products?category`?

miltonhultgren (changes requested):
I think we can make some improvements here!

viedamayanti (dismissed):

shravanipagilla (dismissed):

miltonhultgren:
@ShravyaGurmitkal Are you planning on making more changes in this PR or should I review it based on the last commit? 
If you're not going to address my other comments then I'd appreciate a comment about why! 👍🏼

packages/server/api/routes/products.router.js
ShravyaGurmitkal: Yeah ,but  search and allproducts is done by other team the category is done by our team the whole thing should be managed at some point

packages/server/api/controllers/products.controller.js
ShravyaGurmitkal: First it was the only http error I have changed in the return to object passing the details to become easier when I fetch the data in component

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
ShravyaGurmitkal: If there are 3 it will show three items and when we will click the offset will increase there will be no related data then it will display the first three products from all products and disables the forward button

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
ShravyaGurmitkal: Here if there are less than 3 I am showing the first three products from all products and disabling the forward button

ShravyaGurmitkal: Are you saying to use await near fetchAllproducts

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
ShravyaGurmitkal: I didn't understand the comment on useCallback you have mentioned.

Before api/products?category return format  is similar to api/products but I have changed it while doing this component after fetching

ShravyaGurmitkal:
@miltonhultgren I have addressed the comments and if needed i will do more changes.

packages/server/api/controllers/products.controller.js
miltonhultgren: Right, but it would be best if every endpoint acted the same way when an error happens.

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
miltonhultgren: As a user, I don't think I would understand that those next 3 are **not** related items. I think it would be much better if we had a way to count how many related items we have and act on that instead of some fallback behaviour.

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
miltonhultgren: Yes.

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
miltonhultgren: I understand that you changed it, but my concern is that the change creates a problem. 
The code now has to **know** if it's using the `category` query or not. That's not how things generally are done. The API should keep a consistent return format no matter what query options I send it. 

This is my main concern with this change, that we create a split **inside** the endpoint rather than handling the category logic inside the handler no matter if the query is passed in or not. 

Else, I would argue that these should be two **different** endpoints because their code paths are **not** related and their return format is **not** related.

shpomp (changes requested):
-

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
ShravyaGurmitkal: I will change the return format of api/products?category? to same as other endpoints

ShravyaGurmitkal: As you suggested like this "Instead, it would be great if your API actually returned a count as well. So the response could look like:{
  status: 200,
  total: 13,
  data: [
    product1,
    product2,
    product3
  ].
}
Can I return when its success

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
```diff
@@ -1,33 +1,77 @@
...
+      } else if (categoryProductData.length < 3) {
+        setHasMoreProducts(false);
+        setCategoryProducts(categoryProductData);
+      } else {
+        setHasMoreProducts(true);
+        setCategoryProducts(categoryProductData);
+      }
+    }
+    async function fetchAllProducts() {
+      const response = await fetch(`${apiURL()}/products`);
```
miltonhultgren, line 41:
```suggestion
      const response = await fetch(`${apiURL()}/products?limit=3`);
```

If I look at the URL above (`api/products?category=${category}&limit=3&offset=${offSet}`) then I would assume that I could send limit here as well rather than using `slice`.
If that's not possible, we should change it so that it **is** possible because query parameters should not vary like that.

miltonhultgren (dismissed):

packages/client/src/containers/RelatedItems/RelatedItems.Container.js
ShravyaGurmitkal: @miltonhultgren yes in the endpoint for getAllproducts we don't have any limit query parameter so I haven't used there and used slice

ShravyaGurmitkal: We will update it when the endpoint is updated but for now for this PR I think to do. with slice

shpomp:
I broke it by rebasing. 
@ShravyaGurmitkal will find out what happened and/or needs to happen with ViewAfterCarousel 🤞🏽

shpomp (approved):

shravanipagilla (approved):

PR #86, opened 2022-12-07
feature(variants-endpoint)Get all variants by productId

shravanipagilla:
# Description:
variants-endpoints by productId

# Jira:


packages/server/api/routes/variants.router.js
```diff
@@ -0,0 +1,39 @@
...
+ *        description: Unexpected error.
+ */
+router.get('/:productId', (req, res, next) => {
+  variantsController
+    .getVariantsByProductId(req.params.productId)
+    .then((result) => res.json(result))
+    .catch(next);
+});
+
+module.exports = router;
```
ShravyaGurmitkal, line 41: 👍

ShravyaGurmitkal (dismissed):

packages/server/migrations/20221206115201_add_variants_table.js
```diff
@@ -20,7 +20,7 @@ exports.up = function (knex) {
       'orange',
     ]);
     table.enum('size', ['XS', 'Small', 'Medium', 'Large', 'XL', 'XXL', '3XL']);
-    table.binary('picture');
+    table.string('picture');
```
miltonhultgren: Might wanna make this `pictureUrl`

packages/server/api/routes/variants.router.js
```diff
@@ -0,0 +1,39 @@
...
+ * /variants/{productId}:
+ *  get:
+ *    tags:
+ *    - variants
+ *    summary: Get variants by productId
+ *    description:
+ *      Will return All variants with a matching productId.
+ *    produces: application/json
+ *    parameters:
+ *     - in: path
```
miltonhultgren: I don't think this should be a path variable, I'd rather see it as a query parameter. 

Path segments are meant to express **resources**. So `/users` is the list of all user resources, `/users/{id}` gives you a specific User resource based on it's ID. 

So here, `/variants` gives you a list of all variants, and `/variants/{id}` should give me the variant with that ID but that's not what you mean here. 

You mean to **filter** the list of `variants` to only those that have the matching product ID. So I would say `/variants?product={ID}` instead.

packages/server/api/controllers/variants.controller.js
```diff
@@ -0,0 +1,24 @@
+const knex = require('../../config/db');
+const HttpError = require('../lib/utils/http-error');
+
+const getVariantsByProductId = async (productId) => {
+  if (!productId) {
```
miltonhultgren: This won't really work, what if I pass in 0. 0 is a number but it's falsy so this check will throw. 
Also note that in HTTP all parameters are strings, so you need to convert it to a number first. 

Roughly, you should combine a "not empty" check with a `parseInt` check.

packages/server/api/controllers/variants.controller.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    const variants = await knex('variants')
+      .select('*')
+      .where('product_id', `${productId}`);
+    if (variants.length === 0) {
+      throw new HttpError(`incorrect entry with the id of ${productId}`, 404);
+    }
+    return variants;
+  } catch (error) {
+    return error.message;
```
miltonhultgren: By having this catch here, you remove the point of using HttpError, since it will catch that an unwrap it again. 

You could check if the `error` is a instance of `HttpError` and act based on that.

miltonhultgren (dismissed):

packages/server/migrations/20221206115201_add_variants_table.js
shravanipagilla: @miltonhultgren can you explain a bit more? i did not see any datatype called pictureUrl.

miltonhultgren: I meant, `table.string('pictureUrl);` since what you will store there is the URL

viedamayanti (approved):

shpomp (approved):

RZhyvitskyi (approved):

packages/server/seeds/development/variants.js
```diff
@@ -9,118 +9,136 @@ exports.seed = async function (knex) {
   await knex('variants').insert([
     {
       color: 'red',
-      picture: '.jpg',
+      pictureUrl: 'packages/client/src/assets/jeans&shoes.jpg',
```
miltonhultgren, line 12: @shpomp @zkwsk  I don't know if/how this will work 🤔 

When doing build, it will copy the assets from `src/assets` **if** they are imported by the code, and it will copy whatever is in the `public` folder. 
Since these URLs here might **not** be imported by the JS code, we have to make sure they're available in `public` and that the URLs used here match what it will be at build time, but I don't know at all how that will work at dev time...

I don't know if `create-react-app` actually accounts for this or if they assume that all assets are bundled with the code.

packages/server/api/routes/variants.router.js
```diff
@@ -0,0 +1,41 @@
...
+ *      5XX:
+ *        description: Unexpected error.
+ */
+router.get('/', (req, res, next) => {
+  if ('product' in req.query) {
+    variantsController
+      .getVariantsByProductId(req.query.product)
+      .then((result) => res.json(result))
+      .catch(next);
+  }
```
miltonhultgren, line 38: I know this PR is already merged but this is a pattern we should really avoid.

We should not create branches based on if a parameter is present or not, we should write our code as if all parameters were sent and instead provide good defaults. If the parameter is required but no good default can be found, the API should throw an error instead.

PR #87, opened 2022-12-07
feature (userContext): with firebase

andrii1:
# Description:

# Jira:


Hailegebriel22 (dismissed):

shpomp (dismissed):

packages/client/src/components/Navigation/Navigation.component.js
```diff
@@ -1,30 +1,11 @@
...
-    } catch (err) {
-      /* error */
-    }
-  }, [user?.uid]);
-  useEffect(() => {
-    if (user) {
-      fetchUserName();
-    }
-  }, [user, fetchUserName]);
+  const { user, name, logout } = useUser();
```
miltonhultgren: I think it would make more sense to name the custom hook `useUserContext`. Being aware that this is a context based value can be helpful.

packages/client/src/userContext.js
```diff
@@ -0,0 +1,70 @@
...
+  auth,
+  db,
+  logInWithEmailAndPassword,
+  signInWithGoogle,
+  registerWithEmailAndPassword,
+  logout,
+} from './firebase';
+import { query, collection, getDocs, where } from 'firebase/firestore';
+
+export const UserContext = createContext();
```
miltonhultgren, line 21: Nitpick: Not sure you really want to export this, since the gain of custom hook and custom provider is to not expose this detail.

packages/client/src/userContext.js
```diff
@@ -0,0 +1,70 @@
...
+    <UserContext.Provider value={fireBaseProviderValue}>
+      {children}
+    </UserContext.Provider>
+  );
+}
+
+export function useUser() {
+  const value = useContext(UserContext);
+  return value;
+}
```
miltonhultgren: Here you'd also want to check if `value` is actually filled out. 
If a user has failed to wrap the calling component in your provider they won't have the context and this should fail.

packages/client/src/userContext.js
```diff
@@ -0,0 +1,70 @@
...
+
+UserProvider.propTypes = {
+  children: PropTypes.shape.isRequired,
```
miltonhultgren: children: PropTypes.oneOfType([
        PropTypes.arrayOf(PropTypes.node),
        PropTypes.node
    ]).isRequired

packages/client/src/components/Navigation/Navigation.component.js
andrii1: changed it in Navigation, Login, Signup components.

packages/client/src/userContext.js
andrii1: added

packages/client/src/userContext.js
andrii1: I added this:
`export function useUserContext() {
  const value = useContext(UserContext);
  if (!value) {
    throw new Error('Did you forget to put this under UserProvider');
  }
  return value;
}`

shpomp (changes requested):
-

Hanouj (approved):

shpomp (approved):

PR #88, opened 2022-12-08
feat(single form validation function) for login/contact us page

Hailegebriel22:
# Description:

Extract the previous two form validation functions into one function.

# Jira:


miltonhultgren (changes requested):
Great that you're refactoring to only keep one copy of the same logic! 

I would like though to rename a few things. `FormValidation` sounds more like the name of a component than a function. How about `validateForm` instead?

miltonhultgren (approved):
🚀

shpomp (approved):

PR #90, opened 2022-12-08
Class22 105 get all reviews and get reviews by product id endpoint

ShravyaGurmitkal:
# Description:

Reviews router has two routes one will get all reviews and other one will get reviews by productID

# Jira:

https://hackyourfuture-dk.atlassian.net/browse/CLASS22-105


packages/server/api/routes/reviews.router.js
```diff
@@ -0,0 +1,62 @@
...
+ *  get:
+ *    tags:
+ *    - reviews
+ *    summary: Get all reviews
+ *    description:
+ *      Will return all reviews.
+ *    produces: application/json
+ *    responses:
+ *      200:
+ *        description: Successful request
```
miltonhultgren: Please describe the shape of the returned data. Is it a list? What properties do the items in that list have?

packages/server/api/routes/reviews.router.js
```diff
@@ -0,0 +1,62 @@
...
+router.get('/', (req, res, next) => {
+  reviewsController
+    .getAllReviews()
+    .then((result) => res.json(result))
+    .catch(next);
+});
+
+/**
+ * @swagger
+ * /reviews/{productId}:
```
miltonhultgren: I don't think this should be a path variable, I'd rather see it as a query parameter. 

Path segments are meant to express **resources**. So `/users` is the list of all user resources, `/users/{id}` gives you a specific User resource based on it's ID. 

So here, `/reviews` gives you a list of all reviews, and `/reviews/{id}` should give me the review with that ID but that's not what you mean here. 

You mean to **filter** the list of `reviews` to only those that have the matching product ID. So I would say `/reviews?product={ID}` instead.

packages/server/api/controllers/reviews.controller.js
```diff
@@ -0,0 +1,38 @@
...
+const getAllReviews = async () => {
+  try {
+    const allReviews = await knex.select().table('reviews');
+
+    if (allReviews.length === 0) {
+      throw new HttpError(`No reviews`, 404);
+    }
+    return allReviews;
+  } catch (error) {
+    return error.message;
```
miltonhultgren: By having this catch here, you remove the point of using HttpError, since it will catch that an unwrap it again. 

You could check if the `error` is a instance of `HttpError` and act based on that.

packages/server/api/controllers/reviews.controller.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+};
+
+const getReviewsByProductId = async (productId) => {
+  if (!productId) {
```
miltonhultgren: This won't really work, what if I pass in 0. 0 is a number but it's falsy so this check will throw. 
Also note that in HTTP all parameters are strings, so you need to convert it to a number first. 

Roughly, you should combine a "not empty" check with a `parseInt` check.

miltonhultgren (changes requested):

RZhyvitskyi (dismissed):

packages/server/api/routes/reviews.router.js
```diff
@@ -0,0 +1,44 @@
...
+ *      200:
+ *        description: Successful request
+ *      5XX:
+ *        description: Unexpected error.
+ */
+router.get('/', (req, res, next) => {
+  if (req.query.productId) {
+    reviewsController
+      .getReviewsByProductId(req.query.productId)
+      .then((result) => res.json(result))
```
orhantoy, line 35: Looks like if the `productId` query parameter is set, this route will attempt to send 2 responses. The first one here and the next one on line 40. We can't send 2 responses so the second call will fail and show something in the terminal. Can you try this out and fix?

packages/server/api/controllers/reviews.controller.js
```diff
@@ -0,0 +1,42 @@
+const knex = require('../../config/db');
+const HttpError = require('../lib/utils/http-error');
+
+const getAllReviews = async () => {
+  try {
+    const allReviews = await knex.select().table('reviews');
```
orhantoy, line 6: When would we need to get all reviews? Wouldn't it always be scoped per product, like the other function does?

ShravyaGurmitkal: Can you explain this bit more @orhantoy

miltonhultgren: This function gets **all** reviews, not just the reviews for **one** product (like the other handler in the controller).

ShravyaGurmitkal: Yeah it will return all the reviews like the all products endpoint. Actually we don't need Allreviews endpoint we need only the getreviewsbyproductId endpoint. If you say I can remove that allReviews endpoint

shpomp (dismissed):
As per my understanding, we would just need to describe the response shape and maybe remove getAll in the future - in regards to what was fixed and what was not. 

A lot of good effort to develop and fix things here @ShravyaGurmitkal 💪🏽 
From your replies I also believe you understood and learned something new by the fixes as per review comments 🙏🏽 
Let's merge now and keep in mind that some small changes might still come.

viedamayanti (dismissed):

ShravyaGurmitkal:
@shpomp I have learnt new things while fixing the issues. Yeah later if needed changes will do it.
But now migration is failed I can't merge. Need to check on it.

PR #91, opened 2022-12-09
feature(Modal): new generic component

viedamayanti:
# Description:

# Jira:

https://hackyourfuture-dk.atlassian.net/browse/CLASS22-98

### DIFF
--- a/packages/client/src/components/ProductReviews/ProductReviews.component.js
+++ b/packages/client/src/components/ProductReviews/ProductReviews.component.js
@@ -2,32 +2,29 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { DisplayReviews } from './DisplayReviews/DisplayReviews.component';
 import './ProductReviews.styles.css';
+import Modal from '../Modal/Modal.component';
 
 export const ProductReviews = ({ AllReviews }) => {
-  const [reviews, setReviews] = useState([AllReviews[0]]);
+  const [openModal, setOpenModal] = useState(false);
 
-  const handleMoreReviews = () => {
-    if (reviews.length !== AllReviews.length) {
-      const reviewsIndex = reviews.length;
-      setReviews(reviews.concat(AllReviews[reviewsIndex]));
-    }
+  const toggleModal = () => {
+    setOpenModal(false);
+    document.body.style.overflow = 'visible';
   };
 
   return (
     <div className="product-reviews-wrapper">
-      <h1>Reviews ({AllReviews.length})</h1>
+      <h1 onClick={() => setOpenModal(true)} role="presentation">
+        Reviews ({AllReviews.length})
+      </h1>
 
-      <DisplayReviews reviews={reviews} />
+      <DisplayReviews review={AllReviews[0]} />
 
-      {reviews.length !== AllReviews.length && (
-        <button
-          type="button"
-          onClick={handleMoreReviews}
-          className="readMore-reviews"
-        >
-          Read more
-        </button>
-      )}
+      <Modal title="Reviews" open={openModal} toggle={toggleModal}>
+        {AllReviews.map((review) => {
+          return <DisplayReviews review={review} />;
+        })}
+      </Modal>
     </div>
   );
 };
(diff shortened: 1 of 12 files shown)


packages/client/src/components/ModalComponent/ModalComponent.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './modal.styles.css';
+
+const ModalComponent = ({ handleModal }) => {
```
shpomp: We need some more props to make it really generic and reusable, right? 

So the modal does not know what it needs to show, it just needs to behave right. I think, as a minimum, we would get from props:
modalHeader
modalBody
toggleModal

packages/client/src/components/ModalComponent/ModalComponent.js
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className="overlay">
+      <div className="modal-container">
+        <h5>This is the review pop up</h5>
+        <button
+          type="button"
+          onClick={() => handleModal()}
```
shpomp: would toggleModal() be better?

packages/client/src/components/ModalComponent/ModalComponent.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './modal.styles.css';
+
+const ModalComponent = ({ handleModal }) => {
+  return (
+    <div className="overlay">
+      <div className="modal-container">
+        <h5>This is the review pop up</h5>
```
shpomp: This modal is not just for reviews, so let's start with dynamic body content right away 🙏🏽

packages/client/src/components/ProductReviews/ProductReviews.component.js
```diff
@@ -2,20 +2,31 @@ import React, { useState } from 'react';
...
 
   const handleMoreReviews = () => {
     if (reviews.length !== AllReviews.length) {
       const reviewsIndex = reviews.length;
       setReviews(reviews.concat(AllReviews[reviewsIndex]));
     }
   };
+  const handleModal = () => {
+    setOpenModal(false);
+  };
```
shpomp:
```suggestion
  const toggleModal = () => {
    setShowModal(!showModal);
  };
```

This would be more generic.

shpomp (changes requested):
Very nice 🤩 

A few small fixes to make it really generic and ready to display anything! 


The style will also need a rehaul, which I am sure you know 🙏🏽 
The overlay should be the whole page and the modal should be centered. It should also have min/max height and width, but we can do it the next round!

image

shpomp:
Also, `ModalComponent` can be just `Modal`. 😉

packages/client/src/components/ModalComponent/ModalComponent.js
RZhyvitskyi: I think it's done now

packages/client/src/components/ProductReviews/ProductReviews.component.js
RZhyvitskyi: Done 👍

RZhyvitskyi (dismissed):

RZhyvitskyi:
Mostly it's ready to use for everything.

Have some issue with border.

packages/client/src/containers/ProductReviewContainer/ProductReviewsContainer.js
```diff
@@ -0,0 +1,83 @@
...
+  // useEffect(() => {
+  //   const fetchReviewsByProductId = async () => {
+  //     const response = await fetch(`${apiURL()}/reviews/4`);
+  //     const reviewsProductData = await response.json();
+
+  //     setReviews(reviewsProductData);
+  //   };
+
+  //   fetchReviewsByProductId();
+  // }, [reviews]);
```
shpomp: 🗑️

shpomp (dismissed):

RZhyvitskyi (dismissed):

shravanipagilla (approved):

RZhyvitskyi (approved):

shpomp:
@shravanipagilla @RZhyvitskyi as per common guidelines, you need a peer and a mentor review, which != 2 peers. 👌🏽

PR #92, opened 2022-12-10
feat(FormInput component)

Hailegebriel22:
# Description:

Generic FormInput component. It can be used for text, password, and email types.

# Jira:


packages/client/src/components/Input/FormInput.component.js
```diff
@@ -0,0 +1,47 @@
...
+import validateForm from '../../utils/validateForm';
+import './FormInput.style.css';
+
+const FormInput = ({ value, type, name, label, placeholder, onChange }) => {
+  const [errors, setErrors] = useState({});
+
+  const handleChange = (event) => {
+    setErrors(validateForm(event.target.value));
+    onChange(value);
+  };
```
shpomp: This is kind of a loop, isn't it?
When is the input value ever updated? 
How will you get the value in the component or container that will use the input - to know what to send to the database, f.x.? 
This code, on change, just runs the handle change, but nothing really happens and the input now is no typable. 
I will not go more specific for now, just let me know if it does not make sense and if you don't know what and how to change. 

You are very close though, good job! 💪🏽

shpomp (changes requested):
A component like this definitely needs Storybook, but that is not a priority. 
Create a Jira and a `todo` comment 🙏🏽

packages/client/src/components/Input/FormInput.component.js
Hailegebriel22: i think I need help. I am thinking to use value as a state instead of props. but the mentors suggested it better to use them as props so...

shpomp: yes, you should use it as a prop, and then the state update is also a prop actually! 

```js
// in the parent, when you will want to use the component:
 const [value, setValue] = useState("");
// different names though, as you will have several different inputs. Like name, setName etc. 
...
// the component gets at least one more prop, the state update:
const FormInput = ({ value, setValue, type, name, label, placeholder, onChange }) => { ... 
...
  const handleChange = (event) => {
    setErrors(validateForm(event.target.value));
    setValue(event.target.value);
  };
...
< FormInput value={name} setValue={setName} ....

```

Does this make sense?

shpomp: @Hailegebriel22 ☝🏽

Hailegebriel22: Yes, Thank you so much!

Hanouj (dismissed):

shpomp (changes requested):
I think something went a bit a wrong direction here 🥺 

So the whole idea was to have 1 component, and we discussed the generalising a bit privately too, but now we have 3 components 😬 

You should have 1 FormInput component, that, as we discussed, "becomes" whatever exact input it needs to be by the help of what is passed as props 🙏🏽

Hailegebriel22:
@shpomp 
I did create only one component then 
@miltonhultgren suggested 
"I still recommend creating separate components such as TextField, Select etc."
so Should I make it only one again?

Hailegebriel22:
@miltonhultgren  warning-console
Warning: Invalid value for prop `set value` on <input> tag. Either remove it from the element or pass a string or number value to keep it in the DOM. For details, see https://reactjs.org/link/attribute-behavior.

shpomp:
> @shpomp I did create only one component then @miltonhultgren suggested "I still recommend creating separate components such as TextField, Select etc." so Should I make it only one again?

I just did not know, and all we talked about was 1 generic component. So you could have just made a comment after the commit explaining it here in the PR, so I don't have to wonder.  I was seeing the opposite of what we discussed with no further context and there is a very simple solution to that - communication. 

I don't want to pull you back and forth here, I know how difficult that would be for you, so up to you as a developer to decide 💪🏽 
I do not see a point in 3 - they are absolutely identical. 3 almost identical pieces of code. But you might have discussed more broadly with Milton, I trust you to decide 👍🏽

miltonhultgren:
@shpomp A short summary of our discussion here:
Looking at other solutions like Material UI design system, EUI (Elastic's design system) and the design system I used to work for at my previous job: they all split form controls into individual components.
This is largely because as components grow and evolve they become complex and having so many responsibilities in the same component becomes hard, quickly. It also leads to more risk as a change for "one" variant is likely to break others, and it also makes it harder to work in parallel due to smaller more condensed code surface.
While they might look the same today, with time they are likely to diverge because fundamentally they are different.

packages/client/src/components/Input/EmailFormInput.component.js
```diff
@@ -0,0 +1,65 @@
...
+  placeholder,
+}) => {
+  const [formErrors, setFormErrors] = useState('');
+
+  const handleChange = (e) => {
+    setvalue(e.target.value);
+  };
+
+  const handleValidation = () => {
+    setFormErrors(validateForm(value, name, type));
```
miltonhultgren: I don't think the field itself should call the validation function. 
Rather, it would be best if the controlling code does this.

I think instead the field should be given it's validation/errors as a state instead.

So I think you want props like:
`value`, `errors` and `onChange`.

`onChange` is the callback (instead of `setValue`) and in the controller you can then call the validation and update both `value` and `errors` if needed. The field component should then simply render the error message if needed.

This means the controller should own the state for the error as well. So you might want to consider writing a custom hook for handling those 3 things (value, errors, onChange).

This also aligns more with the idea of doing validation on the `form` itself and the function you call here seems to be for the whole form, not a single field.

packages/client/src/components/Input/EmailFormInput.component.js
```diff
@@ -0,0 +1,65 @@
...
+  setvalue: PropTypes.func,
+  name: PropTypes.string,
+  label: PropTypes.string,
+  placeholder: PropTypes.string,
+  type: PropTypes.string,
+};
+
+EmailFormInput.defaultProps = {
+  value: '',
+  setvalue: '',
```
miltonhultgren: You say above in the propTypes that this is a function but here you say it's a string.

miltonhultgren (changes requested):
I'd also like to rename the folder and components to something like:
FormContorls
  PasswordField
  EmailField
  TextField

miltonhultgren (commented):
@shpomp @Hailegebriel22 

I don't know how we should proceed with this PR.

In general, it tries to do too much at once which makes it hard to review. 

I think the component code that has been added is fine, but the changes around it I'm not so sure about both because;
a) Why are they in this PR?
b) Do they really make sense? 

I'm also a little hesitant if this is really needed, the inputs already exist so why replace them with these ones now? If we had built the shared components first (like in the other FPs) then it would have made more sense. 

I'm trying to see what would create the most learning for @Hailegebriel22 now. Maybe closing this PR and doing something else would be more fun, or to work on how form validation might work and not bother with the shared components.

Hailegebriel22:
@miltonhultgren you suggested that the validation should not be on the components instead it should be on the form, using custom hooks. so I have to rewrite the form validation, and component props accordingly and use the custom hooks to pass the states.

miltonhultgren:
Fair, perhaps I should have been more clear about that such a change didn't need to happen in this PR.

PR #95, opened 2022-12-11
fix(contactUs):validation removed

Hanouj:
# Description:

# Jira:

### DIFF
--- a/packages/client/src/containers/ContactUs.container.js
+++ b/packages/client/src/containers/ContactUs.container.js
@@ -43,7 +43,7 @@ export const ContactUs = () => {
     const { fullname, email, message } = formValues;
     const isInvalid = fullname === '' || email === '' || message === '';
     setDisabled(isInvalid);
-  }, [formValues]);
+  }, [formValues, setDisabled]);
 
   return (
     <main className="contactUs-wrapper">
--- a/packages/client/src/containers/ContactUs.container.js
+++ b/packages/client/src/containers/ContactUs.container.js
@@ -45,25 +45,6 @@ export const ContactUs = () => {
     setDisabled(isInvalid);
   }, [formValues]);
 
-  const validation = (values) => {
-    const errors = {};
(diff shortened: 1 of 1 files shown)


shpomp (changes requested):
What is this? Why? 🤔 
It says validation removed, but it adds something and introduces a build fail.
Jira and PR has zero description 🤷🏽

Hanouj:
> What is this? Why? 🤔 It says validation removed, but it adds something and introduces a build fail. Jira and PR has zero description 🤷🏽

Hanouj:
Hi Karo I just removed the validation because Haile have made a validateForm for both the contectUs and signUp container

packages/client/src/containers/ContactUs.container.js
```diff
@@ -10,8 +10,11 @@ export const ContactUs = () => {
   });
   const [formErrors, setFormErrors] = useState({});
   const [isSubmit, setIsSubmit] = useState(false);
+  const [disabled, setDisabled] = useState(true);
 
   const handleChange = (e) => {
+    setFormErrors({});
+    setIsSubmit(false);
```
miltonhultgren, line 17: I don't think this is needed, the form errors will be set once the form is submitted since that's when the validation happens.

Further, the submit flag should be changed to reflect what you actually mean "messageSentSuccessfully".
Since that will depend on the API, for now it's enough to rename it and only set it to true once the form is submitted.

packages/client/src/containers/ContactUs.container.js
```diff
@@ -34,10 +36,15 @@ export const ContactUs = () => {
     if (Object.keys(formErrors).length === 0 && isSubmit) {
       console.log(formValues);
     }
```
miltonhultgren, line 38: Please remove this effect and remove all the console.log from this file and the eslint disables.

miltonhultgren (changes requested):

PR #98, opened 2022-12-11
feature(Sorting filter Bug and logic): Fix sorting function bug remov…

patel-prakashkumar:
Fix bug from sort function. Remove useEffect from sort function. 

# Description:

# Jira:


miltonhultgren:
Let's work on writing short commit messages. 

"feature(Sorting filter Bug and logic): Fix sorting function bug remov"

So is it a feature or a fix?
Your scope section is too big, it does not need to be so detailed. "Collections" would be enough to tell me which area of the code base this is touching. 

If it **is** a fix, don't say "fix" and try to avoid totally explaining the code in the commit, just explain the how and why, not the what. 

"fix(Collections): Remove sorting effect".

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -54,31 +54,29 @@ export const CollectionsPage = () => {
...
+    }
+    if (sortOrder === 'Price Low to High') {
+      return Number(a.price) > Number(b.price) ? 1 : -1;
+    }
+    if (sortOrder === 'Price High to Low') {
+      return Number(a.price) < Number(b.price) ? 1 : -1;
+    }
+    return 0;
+  }
+  filteredProducts.sort(sortFunction);
```
miltonhultgren: I'm not sure if this is the right thing to do though. 

This way, the list is sorted on **every** render even if the list or the sort order didn't change. The usage of `useEffect` prevents that, why did you feel it was needed to remove it?

patel-prakashkumar: Thank you for feedback. I have tried all way. Its come same result with bug when I used useEffect and I think its come error because after fetching data from API, Its store in products array and products array filter with search input and store with filteredProducts array. filteredProduct store Valued, I used for sorting in function.Thats reason useEffect is not working. If I use useEffect with products Array which is fetch data from API than Its working perfectly than I get problem with search bar Its not working.  I am working on one another solution, I am trying to make sortOrder endpoint where I can sort producs by fetch API.

patel-prakashkumar: Yaah, You are right, I am not write commit comment properly. Now I have cleared for how to write sort message.

miltonhultgren: If you're having issue with many parts of state conflicting that usually means you have to lift more of the state higher up. So that the query string, the products, the sort order, the API call, happen in the same place so that the `useEffect` can react properly. 
Moving the sorting to the API is one solution but I would try first to learn how to keep your state in sync since you'll face similar challenges in the future.

miltonhultgren (approved):

27ManishaPatel (approved):

PR #101, opened 2022-12-14
Class22 119 product cart styling and add delete icon

viedamayanti:
# Description:
style (Product-cart): add delete icon, styling background table-cart, without responsiveness.

# Jira:
https://hackyourfuture-dk.atlassian.net/jira/software/projects/CLASS22/boards/5?selectedIssue=CLASS22-119


nmoskaleva:
Thanks for adding the description and the link to Jira.
You don’t have to mention this doesn’t add functionality, as the PR is clearly for styles. On the other hand, if responsiveness is not yet done, it would make sense to mention it as a to-do. 
Also, since you added styles, an image of before and after could be a great addition to the description. 

Could we improve the commit messages and stick to syntactic guidelines from [here](https://hackyourfuture-cph.gitbook.io/final-project-boilerplate/in-depth/creating-pull-requests)? 
So the structure is: `prefix(Scope): <your short message>`.
For your PR, it would mean something like `style(Cart): add delete icon, …(what else did you add/improve?)`

shpomp (dismissed):

shravanipagilla (approved):

PR #102, opened 2022-12-14
feature(FavoritesPage): add functionality on Favorites page

Yuliia-Balandiuk:
# Description:
Added the ability to create favorite products on the category page. they are displayed on the "favorites" page. The dropdowns will be implemented after refactoring (Karo agreed)

# Jira:
https://hackyourfuture-dk.atlassian.net/jira/software/projects/CLASS22/boards/5?assignee=6054aa834a91ad0068d3cd2d&selectedIssue=CLASS22-97


packages/client/src/components/ProductList/ProductList.component.js
```diff
@@ -3,11 +3,25 @@ import PropTypes from 'prop-types';
...
     return (
       <div className="product">
-        <ProductCard title={product.name} price={product.price} />
+        <ProductCard
+          key={product.id}
+          title={product.name}
+          price={product.price}
+          id={product.id}
+          toggleFavorite={toggleFavorite}
+          isFavorite={favoriteProducts.some((x) => x.id === product.id)}
```
miltonhultgren, line 22: For a small set of products this kind of code is probably fine, but if you have a lot of favorites this could become a problem since it will do this check logic on each render for each product for each favorite. 
This is why it's often better to figure out a way to join this kind of data as early as possible so it becomes one piece of data instead of something that needs to be "linked" on the fly every time it's needed.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -4,6 +4,9 @@ import './CollectionsPage.style.css';
 import DropDownView from '../../components/CategoriesListDropDown/CategoriesListDropDown.component';
 import SearchInput from '../../components/SearchInput/SearchInput.component';
 import { apiURL } from '../../apiURL';
+import { getLocalStorage, setLocalStorage } from '../../utils/storageHelpers';
+
+const favoriteProductsStorageKey = 'favorite_products';
```
miltonhultgren, line 9: You could make it more clear that this is a constant by making it CONSTANT_CASE.

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -17,6 +17,8 @@ function Signup() {
   useEffect(() => {
     if (loading) return;
     if (user) navigate('/');
+    /* eslint-disable no-console */
+    console.log(user?.uid);
```
miltonhultgren: Please clean this up :)

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -14,6 +17,10 @@ export const CollectionsPage = () => {
   const [categories, setCategories] = useState([]);
   const [searchInput, setSearchInput] = useState('');
   const [filteredProducts, setFilteredProducts] = useState([]);
+  const [isFavorite, setIsFavorite] = useState(false);
```
miltonhultgren, line 20: Is this really needed? You check this per product later on while this seems to only apply to 1 product?

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -14,6 +17,10 @@ export const CollectionsPage = () => {
...
+  const [favoriteProducts, setFavoriteProducts] = useState(
+    getLocalStorage(favoriteProductsStorageKey) || [],
+  );
```
miltonhultgren, line 23: Consider writing a custom hook to wrap this kind of logic instead, something like `useLocalState`.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -91,6 +98,34 @@ export const CollectionsPage = () => {
     setFilteredProducts(filterResults);
   }, [searchInput, products]);
 
+  const toggleFavorite = (id, title, price, event) => {
```
miltonhultgren, line 99: You could put this into a custom hook as well since it's not really part of the collections or favorite page. This hook could in turn use the `useLocalState` hook.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -91,6 +98,34 @@ export const CollectionsPage = () => {
...
+
+    const inFavoriteList = favoriteProducts.some((e) => {
+      return e.id === id;
+    });
+
+    if (!inFavoriteList) {
+      setFavoriteProducts((product) => {
+        const newList = [...product, newFavoriteItem];
+        setLocalStorage(favoriteProductsStorageKey, newList);
+        return newList;
```
miltonhultgren, line 114: Nitpick: This return doesn't do anything.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -91,6 +98,34 @@ export const CollectionsPage = () => {
...
+      id,
+      title,
+      price,
+    };
+
+    const inFavoriteList = favoriteProducts.some((e) => {
+      return e.id === id;
+    });
+
+    if (!inFavoriteList) {
```
miltonhultgren, line 110: I think this would be better if you used the `if` block to only decide about adding or removing the favorite and saving either path into a variable called `newFavorites`, then you'd only have one call to `setFavoriteProducts`and `setLocalStorage` (which again, could be hidden as a single custom hook).

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
```diff
@@ -1,6 +1,55 @@
-import React from 'react';
+import React, { useState, useMemo } from 'react';
 import SearchInput from '../../components/SearchInput/SearchInput.component';
+import { getLocalStorage, setLocalStorage } from '../../utils/storageHelpers';
+import ProductCard from '../../components/ProductCard/ProductCard.component';
+import './FavoritesPage.style.css';
 
 export const FavoritesPage = () => {
-  return <SearchInput />;
+  const favoriteProductsStorageKey = 'favorite_products';
```
miltonhultgren, line 8: This is the same constant as in the collections page, I'd move it into a new file and use the same constant in both places (or again, a custom hook for favorites :D).

These kinds of cross cutting concerns is where custom hook and contexts are great.

miltonhultgren (changes requested):
Just curious, but is there a reason why we're storing this only in the client and not in the database for now?

For this kind of solution, you might also consider using a context to make this data available to more pages without having to pass all the props through, together with the function to toggle what is a favorite.

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
Yuliia-Balandiuk: Hi Milton. Thanks for the review.
There's really no reason to do this on the client side. Also, thanks for the tip about the custom hook. Yesterday, Caro and I decided to leave everything as it is, so as not to waste time.
In a new task, we will fetch data from our database.

PR #105, opened 2022-12-17
footer at the bottom of search page

Hailegebriel22:
# Description:
Placing the footer at the bottom of the search page

# Jira:
~~https://hackyourfuture-dk.atlassian.net/browse/CLASS22-125~~

https://hackyourfuture-dk.atlassian.net/browse/CLASS22-93


shpomp (changes requested):
As we discussed, the fix is needed to make sure the Footer appears at the bottom of any page. 
The positioning should concern just the Footer itself. 🙏🏽 

You should also never target and style HTML tags - it might select and style something that you don't intent to be styled that way.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -90,7 +90,7 @@ export const CollectionsPage = () => {
   }, [searchInput, products, sortOrder]);
 
   return (
-    <>
+    <body>
```
miltonhultgren: This is a big no-no, don't put a `body` tag anywhere inside of your React components since the React app is rendered **into** the `body` and there can only be one body per page.

shpomp (changes requested):
Once again, the task is to make sure that Footer is always where it needs to be - regardless the page and the size of its content. 

image


image

PR #106, opened 2022-12-17
feature(shoppingCart): adds shopping cart context

RZhyvitskyi:
# Description:

Context works in all places. 

I have only one issue with deleting `order items` from cart if some of the orders is already selected.

# Jira:

https://hackyourfuture-dk.atlassian.net/jira/software/projects/CLASS22/boards/5?selectedIssue=CLASS22-132

### DIFF
--- a/packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
+++ b/packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
@@ -21,7 +21,7 @@ export const DropdownGroup = ({ productId }) => {
     };
 
     fetchVariantsByProductId();
-  }, [productId, variants]);
+  }, [productId]);
 
   variants.forEach((variant) => {
     colors.push(variant.color);
--- a/packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
+++ b/packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
@@ -1,0 +1,91 @@
+import React, {
+  useState,
+  useContext,
+  useCallback,
+  createContext,
+  useMemo,
+  useEffect,
+} from 'react';
+import PropTypes from 'prop-types';
+import { useUserContext } from '../../userContext';
+import { apiURL } from '../../apiURL';
+
+const ShoppingCartContext = createContext({});
+
+export const useShoppingCartContext = () => {
+  return useContext(ShoppingCartContext);
+};
+
+const ShoppingCartProvider = ({ children }) => {
+  const { user } = useUserContext();
+  const [orderItems, setOrderItems] = useState([]);
+  const [action, setAction] = useState(true);
+
+  useEffect(() => {
+    const fetchOrderItems = async (userUid) => {
+      if (action) {
+        const requestBody = {
+          method: 'GET',
+          headers: {
+            token: `token ${user.uid}`,
+            'Content-Type': 'application/json',
+          },
+        };
+
+        const response = await fetch(`${apiURL()}/orders`, requestBody);
+        const orderItemsData = await response.json();
+
+        if (typeof orderItemsData === 'string') return;
+
+        setOrderItems(orderItemsData);
+        setAction(false);
+      }
+    };
+
+    if (user) fetchOrderItems(user.uid);
+  }, [user, action]);
+
+  const removeOrderItem = useCallback(
+    async (product) => {
+      const request = {
+        method: 'DELETE',
+        headers: {
+          authorization: `token ${user.uid}`,
+        },
+      };
+
+      const response = await fetch(
+        `${apiURL()}/orders?orderId=${product.order_id}&variantId=${
+          product.variant_id
+        }`,
+        request,
+      );
+      const productData = await response.json();
+
+      if (typeof productData === 'number') {
+        setAction(true);
+      }
+    },
+    [user],
+  );
+
+  const orderItemsProviderValue = useMemo(
+    () => ({
+      orderItems,
+      removeOrderItem,
+    }),
+    [orderItems, removeOrderItem],
+  );
+
+  return (
+    <ShoppingCartContext.Provider value={orderItemsProviderValue}>
+      {children}
+    </ShoppingCartContext.Provider>
+  );
+};
+
+export default ShoppingCartProvider;
+
+ShoppingCartProvider.propTypes = {
+  children: PropTypes.node.isRequired,
+};
--- a/packages/client/src/App.js
+++ b/packages/client/src/App.js
@@ -15,27 +15,30 @@ import { ContactUs } from './containers/ContactUs.container';
 import { CollectionsPage } from './containers/CollectionsPage/CollectionsPage.Container';
 import { ShoppingCartPage } from './containers/ShoppingCartPage/ShoppingCartPage.container';
 import { FavoritesPage } from './containers/FavoritesPage/FavoritesPage.container';
+import ShoppingCartProvider from './utils/ShoppingCartContext/ShoppingCartContext';
 
 function App() {
   return (
     <div className="app">
       <Router>
         <UserProvider>
-          <Navigation />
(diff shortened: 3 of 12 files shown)


packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
```diff
@@ -14,11 +14,12 @@ export const DropdownGroup = ({ productId }) => {
     const fetchVariantsByProductId = async () => {
       const response = await fetch(`${apiURL()}/variants?product=${productId}`);
       const variantsProductData = await response.json();
+
       setVariants(variantsProductData);
     };
 
     fetchVariantsByProductId();
-  }, [productId, variants]);
+  }, [productId]);
```
miltonhultgren, line 24: Why was this changed?

packages/client/src/components/ViewPageButton/ViewPageButton.component.js
```diff
@@ -2,9 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import './ViewPageButton.styles.css';
 
-export const ViewPageButton = ({ label, backgroundColor }) => {
+export const ViewPageButton = ({ label, backgroundColor, onClick }) => {
```
miltonhultgren: I'm not sure I understand why this is needed?

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
```diff
@@ -0,0 +1,64 @@
...
+  useCallback,
+  createContext,
+} from 'react';
+import PropTypes from 'prop-types';
+import { useUserContext } from '../../userContext';
+
+const ShoppingCartContext = createContext({});
+
+export const useShoppingCart = () => {
+  return useContext(ShoppingCartContext);
```
miltonhultgren, line 16: Here you should check if the context actually has a value because it might not have been wrapped properly.

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
```diff
@@ -0,0 +1,64 @@
...
+      setPendingItems(updatedPendingItems);
+    },
+    [pendingItems, user?.uid],
+  );
+
+  const shoppingCartProviderValue = useMemo(
+    () => ({
+      pendingItems,
+      deletePendingItem,
+      addPendingItem,
```
miltonhultgren: What do you mean by `pending`? The word confuses me :D You mean in the sense that they have yet to be ordered and paid for? 
I would remove that word because I think it adds more confusion than it adds clarity. The fact that they are in the cart already tells me they have not yet been ordered.

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const useShoppingCart = () => {
+  return useContext(ShoppingCartContext);
+};
+
+const ShoppingCartProvider = ({ children }) => {
+  const { user } = useUserContext();
+
+  const [pendingItems, setPendingItems] = useState(() => {
+    const items = localStorage.getItem(user?.uid);
```
miltonhultgren: Maybe you can sync up with @Yuliia-Balandiuk to talk about creating a custom hook for local storage.

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
```diff
@@ -0,0 +1,64 @@
...
+    (itemId) => {
+      const filteredItems = pendingItems.filter(({ id }) => id !== itemId);
+      localStorage.setItem(user?.uid, JSON.stringify(filteredItems));
+      setPendingItems(filteredItems);
+    },
+    [pendingItems, user?.uid],
+  );
+
+  const addPendingItem = useCallback(
+    (test) => {
```
miltonhultgren: Maybe a better name than test here? :)

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
```diff
@@ -0,0 +1,64 @@
...
+      const updatedPendingItems = { ...pendingItems, ...test };
```
miltonhultgren: `pendingItems` is a list, but here you spread it into an object, that's probably a mistake.
Further, `test` is probably and object representing a product so I don't think you want to spread that apart.

miltonhultgren (dismissed):

packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
RZhyvitskyi: It was causing an infinite loop

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
RZhyvitskyi: We have changed mind about ShoppingCartContext, it will be handled with database

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
RZhyvitskyi: Hahaha, true 👍

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
RZhyvitskyi: Make sense 👍 I'll change it to orderItems soon

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
miltonhultgren: You'll put the users current shopping cart into the database as they are preparing it?
That doesn't make a lot of sense in my opinion. Imagine if they stop midway before making an order, you'll end up with a lot of junk in the database. 

For this kind of data, local storage is perfect. The actual order should of course end up in the database once it's confirmed and paid for.

RZhyvitskyi: If we look on web shops which provides best service, they keep cart data in database, Amazon one of them. Space for data is very cheap in our days. So I'm vouting for database cart data.

RZhyvitskyi: But you are right, we do need to use localstorage for unauthorized users

shpomp:
👀

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
miltonhultgren: Space is cheap, but it's mostly cheap **at scale**. There is a million and one things Amazon does that do not apply to this application. 

While I can agree with your point of "best service", there is a lot more that goes into that. Amazon has much more complex systems in place to manage that data, such as expiry dates on incomplete orders, not to mention the amount of effort they have put into making their sales funnel streamlined so that a large majority of shopping carts actually get turned into orders. 

It's good to be inspired, but it is critical to consider how your situation is different from that of the big tech companies, at least until you actually work for one of the big tech companies. Amazon doesn't use MySQL to store this kind of data and at the scale your solving for right now, you don't need such complexity, and in some situations you cannot even afford such complexity. Amazon did not start out huge, they grew!

viedamayanti (dismissed):

ShravyaGurmitkal (dismissed):

viedamayanti (approved):

packages/client/src/components/ViewPageButton/ViewPageButton.component.js
RZhyvitskyi: We can send a onClick handler function.

PR #107, opened 2022-12-18
Class22 131 feature add partial functionality on shopping cart page

viedamayanti:
# Description:

Add functionality for checkbox, dynamic cart (sum total) and remove btn

# Jira:

https://hackyourfuture-dk.atlassian.net/jira/software/projects/CLASS22/boards/5?selectedIssue=CLASS22-131

### DIFF
--- a/packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
+++ b/packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import './ShoppingCartPage.styles.css';
 import CartContainer from '../../components/Cart/CartContainer/CartContainer.component';
 import CartCount from '../../components/Cart/CartCount/CartCount.component';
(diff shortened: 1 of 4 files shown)


packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
```diff
@@ -10,34 +10,45 @@ import { useNavigate } from 'react-router-dom';
...
   const navigate = useNavigate();
-
-  const products = [
+  const [products, setProducts] = useState([
     {
       id: 1,
       description: 'Low waist Jeans',
       amount: 199.99,
       quantity: 1,
+      selected: false,
```
shpomp: This is unfortunately unsustainable 🙏🏽 
We do not and will not have a property of "selected" in any kind of API response once we fetch the shopping cart content from the DB, so refactoring the the fetch would  break your whole "selected" logic.

shpomp (changes requested):

packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
viedamayanti: Hi Karo yes I already made changes for that

shpomp (approved):

packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
miltonhultgren: Why are we even storing/fetching the shopping cart state from the database?

Orders that are still being made by the user should not live in your database, they should live in local storage or a similar client side temporary storage.

Imagine if 100 customers start making orders by adding items but never finish their orders? Over time you'll get more and more half ready orders in your database to clean up.

shpomp: @miltonhultgren we will go with DB for this project. At some point there was a bit of a clash and discussion about local/db and how to make a unified process without blocking each other, and we landed on the shopping cart living in the db with a specific order status. 
Workarounds 😓

PR #112, opened 2022-12-18
feature(collection): Resolve console error in collection page

patel-prakashkumar:
# Description:
I have resolved some console error in collection page.  

# Jira:


packages/client/src/components/CategoriesListDropDown/CategoriesListDropDown.component.js
```diff
@@ -42,8 +39,8 @@ const DropDownView = ({
 DropDownView.propTypes = {
   options: PropTypes.arrayOf(PropTypes.string).isRequired,
   lable: PropTypes.string.isRequired,
-  onSelect: PropTypes.string.isRequired,
-  showFilterIcon: PropTypes.string.isRequired,
+  onSelect: PropTypes.func.isRequired,
+  showFilterIcon: PropTypes.bool.isRequired,
```
miltonhultgren: Not sure why you set it as `isRequired` when you also have a default value above? 

Decide, either it's optional and the users of this component don't have to care, or it's required and they have to pass it in (fix it where it's missing instead).

packages/client/src/components/CategoriesListDropDown/CategoriesListDropDown.component.js
```diff
@@ -16,9 +16,9 @@ const DropDownView = ({
   };
   const optionList =
     options.length > 0 &&
-    options.map((item, index) => {
+    options.map((item) => {
       return (
-        <option key={index.id} value={item}>
+        <option key={Math.random()} value={item}>
```
miltonhultgren: I think you've misunderstood what `key` does. 

`key` is used by the framework to recycle DOM elements so it has to be **consistent**. Using random here is like pouring coca cola into your car engine...
The issue is that `index` is a number, so `index.id` doesn't exist. Just put back `index`.

packages/client/src/components/ProductList/ProductList.component.js
```diff
@@ -6,7 +6,7 @@ import ProductCard from '../ProductCard/ProductCard.component';
 export default function ProductList({ isLoading, products, filteredProducts }) {
   const ListOfProducts = filteredProducts.map((product) => {
     return (
-      <div className="product">
+      <div className="product" key={Math.random()}>
```
miltonhultgren: Same here, use the `index` instead, if there is nothing better.

packages/client/src/components/ProductList/ProductList.component.js
```diff
@@ -25,6 +25,8 @@ export default function ProductList({ isLoading, products, filteredProducts }) {
 }
 ProductList.propTypes = {
   isLoading: PropTypes.bool.isRequired,
-  products: PropTypes.arrayOf(PropTypes.string).isRequired,
-  filteredProducts: PropTypes.arrayOf(PropTypes.string).isRequired,
+  products: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.object]))
```
miltonhultgren: Here, it would be much better if you used `PropTypes.shape`, like:
```
PropTypes.arrayOf(PropTypes.shape({
  id: PropTypes.number,
  title: PropTypes.string,
})).isRequired
```

Now, you'll likely use the shape for Products in many components since it's a core domain object, so best to create it once and re-use it, like:
```
export const ProductPropType = PropTypes.shape({
  id: PropTypes.number,
  title: PropTypes.string,
});

// Other file
import { ProductPropType } from './prop-types';

PropTypes.arrayOf(ProductPropType).isRequired
```

This also makes the code much more clear!

miltonhultgren (changes requested):

packages/client/src/components/CategoriesListDropDown/CategoriesListDropDown.component.js
patel-prakashkumar: If I can remove console error by replacing index but its showing error while rendering in browser. Do not use Array index in keys. https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

packages/client/src/components/CategoriesListDropDown/CategoriesListDropDown.component.js
patel-prakashkumar: Its show error: propType "onSelect" is not required, but has no corresponding defaultProps declaration. same error in propType "showFilterIcon" is not required, but has no corresponding defaultProps declaration.  I am struggling to pass defaultProps declaration.

packages/client/src/components/CategoriesListDropDown/CategoriesListDropDown.component.js
patel-prakashkumar: Its remove one error by key={item.toString()} and 2nd one remove error by key={product.id} because I found product.id in array.

packages/client/src/components/CategoriesListDropDown/CategoriesListDropDown.component.js
miltonhultgren: https://github.com/HackYourFuture-CPH/fp-class22/blob/develop/packages/client/src/components/Button/Button.component.js#L49-L54

shpomp (dismissed):

Yuliia-Balandiuk (dismissed):

27ManishaPatel (dismissed):

packages/client/src/components/ProductPropType/ProductPropType.component.js
```diff
@@ -0,0 +1,6 @@
+import PropTypes from 'prop-types';
```
miltonhultgren: This is good but I wouldn't put it into `/components` since it's not a component, find some other place to put it please :)

packages/client/src/components/ProductCard/ProductCard.component.js
```diff
@@ -44,5 +44,8 @@ const ProductCard = ({ id, title, price, isFavorite, toggleFavorite }) => {
 ProductCard.propTypes = {
   title: PropTypes.string.isRequired,
   price: PropTypes.string.isRequired,
```
miltonhultgren: This by the way, looks like a Product.

packages/client/src/components/ProductPropType/ProductPropType.component.js
patel-prakashkumar: I have moved to outside component and Its in src directory.

packages/client/src/components/ProductCard/ProductCard.component.js
patel-prakashkumar: I having issue with favorite page prop just merge in develop and now Its has changed props product name, I have mentioned in ProductPropType.js as name and favorite page mention props name as title.

shpomp (approved):

Yuliia-Balandiuk (approved):

27ManishaPatel (approved):

PR #115, opened 2022-12-19
Class22 116 update login page new

andrii1:
# Description:

One signup page

# Jira:


packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -1,62 +1,103 @@
...
+
+  const handleValidation = () => {
+    setFormErrors(validateForm(formValues));
+  };
+
+  const cleanUpValidation = () => {
+    setFormErrors({
+      fullname: '',
+      email: '',
+      password: '',
```
miltonhultgren: Why reset it with the keys while the initial value is just the empty object?

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -1,62 +1,103 @@
...
+
+  const handleSubmit = (event) => {
+    event.preventDefault();
+    if (Object.keys(formErrors).length === 0) {
+      registerWithEmailAndPassword(
+        formValues.fullname,
+        formValues.email,
+        formValues.password,
+      );
+    }
```
miltonhultgren: I recommend getting into the habit of flipping your `if` statements, this is called guarding. 
You find the early path to exit, which thanks to the return allows you to keep the rest of the code at the base indent level making it more readable.

```suggestion
    if (Object.keys(formErrors).length !== 0) {
      return;
      // Probably want to do something more here, like inform the user about why the form was not allowed to be submitted.
    }
    
      registerWithEmailAndPassword(
        formValues.fullname,
        formValues.email,
        formValues.password,
      );
    }
```

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -1,62 +1,103 @@
...
+            />
+            sign up with Google
+          </button>
+          <button type="button" className="facebook-signin-btn">
+            <img
+              className="signin-icons"
+              src="../../assets/icons/facebook.png"
+              alt=" facebooklogo"
+            />
+            sign up with Facebook
```
miltonhultgren: Sign in, rather than up?

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -1,62 +1,103 @@
...
+          </button>
+        </div>
+        <p className="or-text">-OR-</p>
+        <SignupForm
+          handleChange={handleChange}
+          handleValidation={handleValidation}
+          cleanUpValidation={cleanUpValidation}
```
miltonhultgren, line 86: Why is cleanUpValidation a special prop, rather than reseting the `formErrors` prop on submit?

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -1,62 +1,103 @@
...
         />
-        <button type="submit" className="register__btn" onClick={register}>
-          Signup
-        </button>
         <button
           type="submit"
-          className="register__btn register__google"
-          onClick={signInWithGoogle}
+          className="signup-submit-button"
+          onClick={handleSubmit}
```
miltonhultgren, line 93: This submit button won't properly be connected to the form since it's outside of the `form` tag.

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -1,62 +1,103 @@
...
         >
-          Signup with Google
+          Signup
         </button>
-        <div>
-          Already have an account? <Link to="/login">Login</Link> now.
-        </div>
+        <br />
+        <span className="dont-want-signup">
+          Dont want to Sign Up? Continue as a Guest
```
miltonhultgren, line 99: Should this be a link or will that be fixed later?

packages/client/src/containers/Signup/SignupForm.js
```diff
@@ -73,16 +48,39 @@ const SignupForm = () => {
       {formErrors.password && (
         <p className="form-erros">{formErrors.password}</p>
       )}
+      {/*
       <button
-        onClick={handleSubmit}
+        /* onClick={handleSubmit}
+        onClick={register}
```
miltonhultgren, line 54: I think you want this button still here and to have `onSubmit` as a prop.

miltonhultgren (changes requested):
I'm confused, the Login page was deleted but the route for it is updated and the SignUp page is added but no route is configured for it?

packages/client/src/containers/Signup/Signup.container.js
andrii1: should be fixed later..

packages/client/src/containers/Signup/Signup.container.js
andrii1: yeah.. I know.. the issue was that if I put a submit button inside the form component, firebase doesn't work

packages/client/src/containers/Signup/Signup.container.js
andrii1: fixed

packages/client/src/containers/Signup/Signup.container.js
andrii1: hm don't know, I just merged Heile's part.
Changed to this:
`const cleanUpValidation = () => {
    setFormErrors({
      fullname: {},
      email: {},
      password: {},
    });
  };`

miltonhultgren (dismissed):

packages/client/src/containers/Signup/Signup.container.js
miltonhultgren: Then it would be better to debug that issues rather than do some work around.

miltonhultgren (approved):

PR #118, opened 2022-12-20
feature(order_items endpoint): Creates endpoint to fetch order_items

RZhyvitskyi:
# Description:

Now we can use userUid to fetch users order_items

# Jira:

https://hackyourfuture-dk.atlassian.net/jira/software/projects/CLASS22/boards/5?selectedIssue=CLASS22-135

### DIFF
--- a/packages/server/api/controllers/orders.controller.js
+++ b/packages/server/api/controllers/orders.controller.js
@@ -1,0 +1,38 @@
+const knex = require('../../config/db');
+const HttpError = require('../lib/utils/http-error');
+
+const getOrderItemsByUserId = async (token) => {
+  try {
+    if (!token) {
+      throw new HttpError('You are not authenticated', 401);
+    }
+
+    const userUid = token.split(' ')[1];
+
+    if (!userUid) {
+      throw new HttpError('Token is wrong', 400);
+    }
+
+    const user = (await knex('users').where({ uid: userUid }))[0];
+
+    if (!user) {
+      throw new HttpError('User was not found', 404);
+    }
+
+    const orderItems = await knex('orders')
+      .where({
+        user_id: user.id,
+        status: 'created',
+      })
+      .join('order_items', 'orders.id', 'order_items.order_id')
+      .join('variants', 'order_items.variant_id', 'variants.id');
+
+    return orderItems;
+  } catch (error) {
+    return { ...error, message: error.message };
+  }
+};
+
+module.exports = {
+  getOrderItemsByUserId,
+};
--- a/packages/server/api/routes/orders.router.js
+++ b/packages/server/api/routes/orders.router.js
@@ -1,0 +1,47 @@
+const express = require('express');
+
+const router = express.Router({ mergeParams: true });
+
+// controllers
+const ordersController = require('../controllers/orders.controller');
+
+/**
+ * @swagger
+ * /orders:
+ *  get:
+ *    tags:
+ *    - orders
+ *    summary: Get all orders for user
+ *    description:
+ *      Will return an array with all orders for a specific user.
+ *    produces: application/json
+ *    parameters:
+ *     - in: header
+ *       name: token
+ *       schema:
+ *         type: string
+ *         required: true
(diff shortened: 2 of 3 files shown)


packages/server/api/routes/orders.router.js
```diff
@@ -0,0 +1,47 @@
+const express = require('express');
+
+const router = express.Router({ mergeParams: true });
+
+// controllers
```
miltonhultgren, line 5: Nit: This comment doesn't add any value.

packages/server/api/routes/orders.router.js
```diff
@@ -0,0 +1,47 @@
...
+const ordersController = require('../controllers/orders.controller');
+
+/**
+ * @swagger
+ * /orders:
+ *  get:
+ *    tags:
+ *    - orders
+ *    summary: Get all orders for user
```
miltonhultgren, line 14: Isn't this just the order items?

packages/server/api/routes/orders.router.js
```diff
@@ -0,0 +1,47 @@
...
+ *        description: Token is wrong
+ *      401:
+ *        description: You are not authenticated
+ *      404:
+ *        description: User was not found
+ *      5XX:
+ *        description: Unexpected error.
+ */
+router.get('/', (req, res, next) => {
+  const { token } = req.headers;
```
miltonhultgren, line 39: Token isn't a very explanatory name, it makes me think mostly of a security token which this isn't. 
This is simply a `userId` which the controller function calls it.

Further, and perhaps equally important, why is this coming in via headers? I would expect this to be a query parameter since it affects how the endpoint responds. While I would use headers for things like metadata or security data which is a orthogonal concern from how the endpoint behaves.

packages/server/api/controllers/orders.controller.js
```diff
@@ -0,0 +1,38 @@
...
+        user_id: user.id,
+        status: 'created',
+      })
+      .join('order_items', 'orders.id', 'order_items.order_id')
+      .join('variants', 'order_items.variant_id', 'variants.id');
+
+    return orderItems;
+  } catch (error) {
+    return { ...error, message: error.message };
+  }
```
miltonhultgren, line 33: I don't think this try/catch is needed, the router already does a `.catch` so just let these errors throw up into that instead.

packages/server/api/controllers/orders.controller.js
```diff
@@ -0,0 +1,38 @@
...
+
+    const user = (await knex('users').where({ uid: userUid }))[0];
+
+    if (!user) {
+      throw new HttpError('User was not found', 404);
+    }
+
+    const orderItems = await knex('orders')
+      .where({
+        user_id: user.id,
```
miltonhultgren, line 24: Why do we need to first resolve the user to get their ID? Doesn't the endpoint accept the `userId` already? Why would it change in between?

packages/server/api/controllers/orders.controller.js
```diff
@@ -0,0 +1,38 @@
...
+  try {
+    if (!token) {
+      throw new HttpError('You are not authenticated', 401);
+    }
+
+    const userUid = token.split(' ')[1];
+
+    if (!userUid) {
+      throw new HttpError('Token is wrong', 400);
+    }
```
miltonhultgren, line 14: I don't think this code makes much sense in this kind of non-production app. 

While it makes a lot of sense to do server side verification, it only makes sense if you're actually doing that and not just doing code that pretends to do that. 
That said, from a learning perspective, I would look into creating an Express middleware to handle this verification instead so that the feature controller don't have to care, they can assume the request is verified already. 
That way you don't need to duplicate this check in every endpoint.

miltonhultgren (changes requested):
Let's clarify in the code and commit messages if this endpoint pulls out orders or order items for a specific user. While we do join in all the order items, from the users perspective this is just an order so I think it would make sense to just rename things to orders, we don't have to leak the details of the database here.

Further, let's stick to using query parameters which is what the other endpoints use.

PR #119, opened 2022-12-22
Class22 119 product cart styling and add delete icon

viedamayanti:
# Description:

# Jira:


shpomp (changes requested):
What is this PR about? 
The name vs. what files are changed is a mystery, I also don't see any Jira 🥺

viedamayanti:
Hi Karo, its styling for the shopping cart and adding the delete icon for that

PR #120, opened 2022-12-27
Refactor(products endpoint) to consider user and favorites

Yuliia-Balandiuk:
# Description:

Refactor products endpoint to consider user and favorites

# Jira:
https://hackyourfuture-dk.atlassian.net/browse/CLASS22-141


packages/client/src/components/ProductList/ProductList.component.js
```diff
@@ -8,8 +8,7 @@ export default function ProductList({
   isLoading,
   products,
   filteredProducts,
-  toggleFavorite,
-  favoriteProducts,
+  // toggleFavorite,
```
miltonhultgren: Please don't comment out code, delete it if it isn't needed and if you change your mind, Git allows you to revert changes.

packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -17,10 +16,7 @@ export const CollectionsPage = () => {
   const [categories, setCategories] = useState([]);
   const [searchInput, setSearchInput] = useState('');
   const [filteredProducts, setFilteredProducts] = useState([]);
-  const [isFavorite, setIsFavorite] = useState(false);
-  const [favoriteProducts, setFavoriteProducts] = useState(
-    getLocalStorage(favoriteProductsStorageKey) || [],
-  );
+  let isFavorite;
```
miltonhultgren, line 19: This is never assigned any value?

packages/server/api/controllers/products.controller.js
```diff
@@ -2,8 +2,18 @@ const knex = require('../../config/db');
 const HttpError = require('../lib/utils/http-error');
 
 // Get All products
-const getProducts = async () => {
-  return knex.select().table('products');
+const getProducts = async (token) => {
+  const userUid = token.split(' ')[1];
+  const user = (await knex('users').where({ uid: userUid }))[0];
```
miltonhultgren, line 7: If there is no such user found that should probably be an error thrown.

packages/server/api/controllers/users.controllers.js
```diff
@@ -0,0 +1,10 @@
+const knex = require('../../config/db');
+
+// Get All Users
+const getUsers = async () => {
```
miltonhultgren, line 4: Where is this used?

miltonhultgren (changes requested):

27ManishaPatel (dismissed):

PR #121, opened 2022-12-27
Refactor(Favorites get endpoint): add user authorization

Yuliia-Balandiuk:
# Description:
Favorites page:
- if the user is not logged in, we display the message: "You have no favorite products"
- if the user is logged in, we display list of favorite products
 
# Jira:

https://hackyourfuture-dk.atlassian.net/browse/CLASS22-146


miltonhultgren:
I'm not sure how good/clear the guidelines have been around how to name commits/PRs, but here is what I would suggest instead:
In general, a commit should explain what it does in the title, and maybe explain why it does it in the body. The code should explain how.
From someone who is reading the commit history, the fact that this commit was a refactor doesn't really matter. The reason some commit styles include things like feature/fix/chore is because they use the commits to decide the version to use for [semantic versioning](https://semver.org/), if we're not using that, I'd say let's avoid adding more noise.

The scope part is useful though for those reading the PR to know what area of the code this is in, but it should be short and rather explain the feature area since in the changed files list I can see that it's either only server files changed. In this case, it's actually both the frontend and backend changed so the scope is misleading. 

Finally, the PR doesn't actually _add_ authorization, rather it consumes the token for the already authorized user. 
So I'd suggest this as a commit messages: `(favorites) Use Firebase token to load user favorites`

packages/server/api/routes/favorites.router.js
```diff
@@ -27,10 +27,10 @@ const favoritesController = require('../controllers/favorites.controller');
  *        description: Unexpected error.
  */
 router.get('/', (req, res, next) => {
-  const userId = 1;
-  // TO DO : once we will add authentication I will update it
+  const { token } = req.headers;
```
miltonhultgren, line 30: I don't really understand why we are passing the token here since we don't actually use it to verify that the user is authorized for this request. We could just unpack the user ID in the UserContext and send that instead. 

Or at least unpack the userID in the router so that the controller doesn't have to know about that step.

packages/server/api/controllers/favorites.controller.js
```diff
@@ -2,21 +2,25 @@ const knex = require('../../config/db');
 const HttpError = require('../lib/utils/http-error');
 
 // get by user-id
-const getFavoritesByUserId = async (userId) => {
-  if (isNaN(userId)) {
-    throw new HttpError('Id should be a number', 400);
-  }
+const getFavoritesByUserId = async (token) => {
+  const userUid = token.split(' ')[1];
```
miltonhultgren, line 6: What if no token is sent?

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
```diff
@@ -1,41 +1,54 @@
...
+
+      const response = await fetch(url, {
+        headers: {
+          token: `token ${user?.uid}`,
+        },
+      });
+      const favoritesData = await response.json();
+      if (Array.isArray(favoritesData)) {
+        setFavorites(favoritesData);
+      }
```
miltonhultgren, line 26: In which case wouldn't it be an array? In case it's an error? Then we should handle that error instead.

miltonhultgren (changes requested):

PR #123, opened 2023-01-03
feature(Order endpoint) create order endpoint

Yuliia-Balandiuk:
# Description:

Create an order endpoint for fetch orders with status created by user id

# Jira:
https://hackyourfuture-dk.atlassian.net/browse/CLASS22-128

### DIFF
--- a/packages/server/api/controllers/orders.controller.js
+++ b/packages/server/api/controllers/orders.controller.js
@@ -5,9 +5,11 @@ const HttpError = require('../lib/utils/http-error');
 const getOrdersByUserId = async (token) => {
   const userUid = token.split(' ')[1];
   const user = (await knex('users').where({ uid: userUid }))[0];
-
+  if (!user) {
+    throw new HttpError('User not found', 401);
+  }
   try {
-    const favorites = await knex('orders')
+    const orders = await knex('orders')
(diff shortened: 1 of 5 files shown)


packages/server/api/controllers/orders.controller.js
```diff
@@ -0,0 +1,39 @@
+const knex = require('../../config/db');
+const HttpError = require('../lib/utils/http-error');
+
+// get order with status "created" by user-id
+const getOrdersByUserId = async (token) => {
+  const userUid = token.split(' ')[1];
+  const user = (await knex('users').where({ uid: userUid }))[0];
```
miltonhultgren, line 7: What if no such user is found or passed in?

packages/server/api/controllers/orders.controller.js
```diff
@@ -0,0 +1,39 @@
...
+
+  try {
+    const favorites = await knex('orders')
```
miltonhultgren: orders?

miltonhultgren (changes requested):
Isn't this the same work as https://github.com/HackYourFuture-CPH/fp-class22/pull/118?

orhantoy (approved):

27ManishaPatel (approved):

PR #124, opened 2023-01-03
Refactor (Shopping cart): use order endpoint by user id

Yuliia-Balandiuk:
# Description:

Shopping cart:  add fetching order with status "created" from DB by user id

# Jira:

https://hackyourfuture-dk.atlassian.net/browse/CLASS22-147


packages/server/api/controllers/orders.controller.js
```diff
@@ -0,0 +1,39 @@
+const knex = require('../../config/db');
```
miltonhultgren: I'd try to avoid adding the same change in multiple commits/branches.
Either wait for work to be approved and merged, or point your branch to the branch that adds this work instead.

packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
```diff
@@ -51,7 +74,11 @@ export const ShoppingCartPage = () => {
         </div>
         <CartCount productSum={productSum} />
         <CartContainer>
-          <CartTable products={products} />
+          <CartTable
+            products={products}
+            setProducts={setProducts}
```
miltonhultgren: Why is this table able to set the products directly? If the table component is used to add/remove/change quantity/filter then it should expose those events instead. It should not have access to the state that ShoppingCartPage owns.

packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
```diff
@@ -7,37 +7,60 @@ import CartTotal from '../../components/Cart/CartTotal/CartTotal.component';
...
+        }));
+        setProducts(ordersData);
+      }
+    };
+
+    fetchOpenedOrder();
+    setIsLoading(false);
+  }, [user]);
+
+  const productSum = products
```
miltonhultgren: I'd rename this to itemCount instead.

packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
```diff
@@ -7,37 +7,60 @@ import CartTotal from '../../components/Cart/CartTotal/CartTotal.component';
...
+
+      const response = await fetch(url, {
+        headers: {
+          token: `token ${user?.uid}`,
+        },
+      });
+
+      const fetchedOrdersData = await response.json();
+
+      if (Array.isArray(fetchedOrdersData)) {
```
miltonhultgren, line 35: Let's handle the case when an error is thrown so this isn't an array.

packages/client/src/components/Cart/CartTable/CartTable.component.js
```diff
@@ -1,9 +1,27 @@
...
+export default function CartTable({ products, setProducts, isLoading }) {
+  const { user } = useUserContext();
+  const [selected, setSelected] = useState(false);
+
+  const handleChange = (product) => {
+    const productCart = [...products];
+
+    productCart.forEach((row, index) => {
+      if (productCart[index].id === product.id)
+        productCart[index].selected = !productCart[index].selected;
```
miltonhultgren, line 16: This is a little out of scope, but in general, I would try to avoid mixing application state with UI state. 

In most component libraries you'll have a row expressed as this:
```
{
  selected: true,
  data: { /* Properties for a product */ }
}
```

So that you can make a clear separation between what is state for the table render, and what is actually the data to send to the API for example. This row state can safely be owned by the table component and be easily recreated if the `data` part changes.

miltonhultgren (changes requested):
I don't understand why we load a completed order into the shopping cart state.

If the order has yet to be made, I don't think it should live in the database, session or localStorage are the right place for temporary data.

packages/server/api/controllers/orders.controller.js
Yuliia-Balandiuk: It is not my decision to store the data in the database. What do you mean by "filled order"? Endpoint created for order with status "created"

packages/client/src/containers/ShoppingCartPage/ShoppingCartPage.container.js
Yuliia-Balandiuk: I'm not sure if I understood correctly, but previous mentors insisted that not all states be placed in the parent component.

miltonhultgren (dismissed):

PR #125, opened 2023-01-03
Class22 127 remove favorites confirmation via modal

patel-prakashkumar:
# Description:

# Jira:

### DIFF
--- a/packages/client/src/components/ProductCard/ProductCard.component.js
+++ b/packages/client/src/components/ProductCard/ProductCard.component.js
@@ -7,7 +7,15 @@ import StarRateIcon from '../../assets/star-rate-icon.svg';
 import BackgroundImage from '../../assets/jeans&shoes.jpg';
 import PropTypes from 'prop-types';
 
-const ProductCard = ({ id, title, price, isFavorite, toggleFavorite }) => {
+const ProductCard = ({
+  id,
+  title,
+  price,
+  isFavorite,
+  toggleFavorite,
+  toggleModal,
+  openModal,
+}) => {
   return (
     <div className="product-card-wrapper">
       <div className="product-img-wrapper">
@@ -15,13 +23,17 @@ const ProductCard = ({ id, title, price, isFavorite, toggleFavorite }) => {
         <div className="cart-favorite-wrapper">
           <img src={CartIcon} alt="Cart Icon" />
           <button
-            onClick={(event) => toggleFavorite(id, title, price, event)}
+            onClick={(event) => {
+              toggleFavorite(id, title, price, event);
+              toggleModal(openModal);
+            }}
             type="button"
             className="favorite-button"
           >
             <img
               src={isFavorite ? FullFavoriteIcon : EmptyFavoriteIcon}
               alt="Favorite Icon"
+              role="presentation"
             />
           </button>
         </div>
@@ -47,5 +59,7 @@ ProductCard.propTypes = {
   id: PropTypes.number.isRequired,
   isFavorite: PropTypes.bool.isRequired,
   toggleFavorite: PropTypes.func.isRequired,
+  toggleModal: PropTypes.func.isRequired,
+  openModal: PropTypes.bool.isRequired,
 };
 export default ProductCard;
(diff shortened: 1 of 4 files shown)


packages/client/src/components/ProductCard/ProductCard.component.js
```diff
@@ -7,21 +7,33 @@ import StarRateIcon from '../../assets/star-rate-icon.svg';
...
+                ? setOpenModal({ modalStatus: true, favoriteId: id })
+                : toggleFavorite(id, title, price, event);
+            }}
             type="button"
             className="favorite-button"
           >
             <img
               src={isFavorite ? FullFavoriteIcon : EmptyFavoriteIcon}
               alt="Favorite Icon"
+              role="presentation"
```
miltonhultgren: Not sure why you're adding this? I would avoid changing ARIA roles unless you really know what you are doing.

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
```diff
@@ -3,16 +3,26 @@ import SearchInput from '../../components/SearchInput/SearchInput.component';
 import { getLocalStorage, setLocalStorage } from '../../utils/storageHelpers';
 import ProductCard from '../../components/ProductCard/ProductCard.component';
 import './FavoritesPage.style.css';
+import Modal from '../../components/Modal/Modal.component';
+import { ViewPageButton } from '../../components/ViewPageButton/ViewPageButton.component';
 
 export const FavoritesPage = () => {
+  const [openModal, setOpenModal] = useState({
```
miltonhultgren: `modalState` might be a better name here.

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
```diff
@@ -3,16 +3,26 @@ import SearchInput from '../../components/SearchInput/SearchInput.component';
...
+    modalStatus: false,
+    favoriteId: -1,
+  });
+  const toggleModal = () => {
+    setOpenModal({ modalStatus: false, favoriteId: -1 });
+    document.body.style.overflow = 'visible';
```
miltonhultgren: In general you should never modify the CSS of the page directly, instead you should alter this with a CSS class that is added within the React markup instead.

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
```diff
@@ -3,16 +3,26 @@ import SearchInput from '../../components/SearchInput/SearchInput.component';
...
 import './FavoritesPage.style.css';
+import Modal from '../../components/Modal/Modal.component';
+import { ViewPageButton } from '../../components/ViewPageButton/ViewPageButton.component';
 
 export const FavoritesPage = () => {
+  const [openModal, setOpenModal] = useState({
+    modalStatus: false,
+    favoriteId: -1,
+  });
+  const toggleModal = () => {
```
miltonhultgren: closeModal?

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
```diff
@@ -34,16 +43,42 @@ export const FavoritesPage = () => {
...
           id={product.id}
+          setOpenModal={setOpenModal}
           toggleFavorite={toggleFavorite}
           isFavorite={favoriteProducts.some((x) => x.id === product.id)}
         />
       </div>
     );
   });
 
+  const handleModal = (favoriteId) => {
```
miltonhultgren, line 53: removeFavorite?

miltonhultgren (changes requested):

packages/client/src/components/ProductCard/ProductCard.component.js
27ManishaPatel: That is the element for screen readers to understand but it should not be here so I will remove it, Thank you 😊

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
27ManishaPatel: True I can remove it

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
27ManishaPatel: Yes it is closeModel

packages/client/src/containers/FavoritesPage/FavoritesPage.container.js
27ManishaPatel: Yes handleModal will remove it

PR #128, opened 2023-01-08
Feature (Add uid to db)

andrii1:
# Description:
Adding uid to users table

# Jira:


packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -26,12 +32,28 @@ function Signup() {
...
+        full_name: fullname,
+        email: userCreated.email,
+        uid: userCreated?.uid,
+      }),
+    };
+    await fetch(`${apiURL()}/users`, requestOptions)
+      .then((response) => response.json())
+      .then((data) => {
+        /* console.log(data);
+        console.log('user', userCreated, userCreated?.uid); */
```
orhantoy: What should happen here?

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -26,12 +32,28 @@ function Signup() {
...
+    const requestOptions = {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        full_name: fullname,
+        email: userCreated.email,
+        uid: userCreated?.uid,
+      }),
+    };
+    await fetch(`${apiURL()}/users`, requestOptions)
```
orhantoy: Is there a particular reason for mixing await/async with `.then` below?

packages/client/src/containers/Signup/Signup.container.js
```diff
@@ -46,8 +68,12 @@ function Signup() {
   };
   useEffect(() => {
     if (loading) return;
-    if (user) navigate('/');
-  }, [user, loading, navigate]);
+    if (user) {
+      /* console.log('user1', user, user.email); */
```
orhantoy: Let's remember to clean up commented code.

packages/server/api/controllers/users.controller.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+};
+
+// post
+const createUsers = async (body) => {
+  try {
+    await knex('users').insert({
+      full_name: body.full_name,
+      email: body.email,
+      uid: body.uid,
```
orhantoy, line 9: Did you figure out a way to deal with duplicate uids? Couldn't that happen if someone tries to sign up twice with the same email?

packages/server/api/routes/users.router.js
```diff
@@ -27,4 +27,47 @@ router.get('/', (req, res, next) => {
     .catch(next);
```
orhantoy: I don't think this is needed so I would remove this route and the `getUsers` controller function.

packages/server/api/routes/users.router.js
```diff
@@ -27,4 +27,47 @@ router.get('/', (req, res, next) => {
...
+ *    parameters:
+ *      - in: body
+ *        name: users
+ *        description: The user to create.
+ *        schema:
+ *          type: object
+ *          required:
+ *            - full_name
+ *            - email
+ *            - created_at
```
orhantoy: I don't think `created_at` is accepted as a parameter.

packages/server/api/routes/users.router.js
```diff
@@ -27,4 +27,47 @@ router.get('/', (req, res, next) => {
...
+ *            - uid
+ *          properties:
+ *            full_name:
+ *              type: string
+ *            email:
+ *              type: string
+ *            created_at:
+ *              type: date/time
+ *            uid:
+ *              type: date/time
```
orhantoy: I think the type here should be string.

packages/server/api/routes/users.router.js
```diff
@@ -27,4 +27,47 @@ router.get('/', (req, res, next) => {
...
+ *    responses:
+ *      201:
```
orhantoy: Does it actually return a 201 response?

packages/server/api/controllers/users.controller.js
andrii1: it goes through firebase first, so user is not created if someone tries to sign up twice with the same email;

packages/client/src/containers/Signup/Signup.container.js
andrii1: removed it

packages/client/src/containers/Signup/Signup.container.js
andrii1: removed

packages/server/api/routes/users.router.js
andrii1: changed to 200, or should it be removed?

packages/server/api/routes/users.router.js
andrii1: ok, it was just for testing purposes

packages/server/api/routes/users.router.js
orhantoy: It's fine with 200

orhantoy (approved):

PR #132, opened 2023-01-08
feature(order_items): post new order item

RZhyvitskyi:
# Description:
Now we can add new order item to order list for specific user by Uid.

image
image
 
Endpoint makes a lot of validation. There must be a better way how to refactor that, using different middleware but I think it's ok for the first iteration.

# Jira:
https://hackyourfuture-dk.atlassian.net/jira/software/projects/CLASS22/boards/5?selectedIssue=CLASS22-151


ShravyaGurmitkal (approved):

viedamayanti (approved):

packages/server/api/controllers/orders.controller.js
```diff
@@ -33,6 +33,83 @@ const getOrdersByUserId = async (token) => {
...
+      product_id: productId,
+      color,
+      size,
+    });
+
+    if (!variant) {
+      throw new HttpError('Variant is not available', 404);
+    }
+
+    const [productsInQueue] = await knex
```
orhantoy, line 77: Couldn't this query could be simplified to something like this:

```sql
select sum(quantity)
from order_items
where order_id = ? AND variant_id = ?
```

packages/server/api/controllers/orders.controller.js
```diff
@@ -33,6 +33,83 @@ const getOrdersByUserId = async (token) => {
...
+
+    const availableProducts = variant.stock - productsInQueue.quantity;
+
+    if (quantity > availableProducts) {
+      throw new HttpError(
+        'Order products quantity is bigger than stock quantity',
+        400,
+      );
+    } else {
+      await knex('order_items').insert({
```
orhantoy, line 97: If I add the same variant multiple times, I would get multiple order items for the same variant. The behavior I would expect to see is that existing order items are updated. I don't think your approach is incorrect though but just mentioning as a potential improvement.

orhantoy (approved):
Overall looks great 💪 
I left some suggestions for improvements but feel free to ignore them for now.

packages/server/api/controllers/orders.controller.js
RZhyvitskyi: Thank, I though about that also. I have improved `get order items` in other PR. So query just groups all of `order items` with the same `order_id` and `variants_id` together. 

But yes, update is much more easy and maybe lighter.

RZhyvitskyi: I though about that, but so some reasons decided to group order items with the same `order_id` and `variant_id` together in other PR, that is pelated to ShoppinCartContext. Updating must be more efficient.

PR #138, opened 2023-01-10
feature(DropdownGroup.container): Displaying the size and quantity ba…

ShravyaGurmitkal:
…sed on selected color

# Description:

Displaying the size and quantity to only specific range for the selected color.
Also changed the swagger comments of variants endpoint.

# Jira:

https://hackyourfuture-dk.atlassian.net/browse/CLASS22-150


packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
```diff
@@ -33,16 +36,56 @@ export const DropdownGroup = ({ productId }) => {
   sizes = [...new Set(sizes)];
   quantities = [...new Set(quantities)];
 
+  const fetchVariantsByColor = async (color) => {
+    let filteredSizes = [];
+    let filteredQuantities = [];
+    const response = await fetch(
+      `${apiURL()}/variants?color=${color}&product_id=${productId}`,
```
orhantoy, line 45: I would prefer that the endpoint URL is like this: `/api/products/:productId/variants`
The color can still be a query param, that's OK.

packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
```diff
@@ -33,16 +36,56 @@ export const DropdownGroup = ({ productId }) => {
...
+    const filteredVariantsData = await response.json();
+    filteredVariantsData.forEach((variant) => {
+      filteredSizes.push(variant.size);
+      filteredQuantities.push(variant.stock);
+    });
+    filteredSizes = [...new Set(filteredSizes)];
+    filteredQuantities = [...new Set(filteredQuantities)];
+    const maxQuantity = Math.max(...filteredQuantities);
+    if (maxQuantity >= 5) {
+      filteredQuantities = [...Array(10).keys()].map((i) => i + 1);
```
orhantoy: What happens here if `maxQuantity` is 7?

packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
ShravyaGurmitkal: But when we want to find variants using variant id then we will use path parameter, its using productId so I have used it as query parameter

ShravyaGurmitkal: Now should we need to create new endpoint

ShravyaGurmitkal: I have used the endpoint which was already created for this task

packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
ShravyaGurmitkal: Yeah I will change this

viedamayanti (approved):

packages/client/src/containers/DropdownGroup/DropdownGroup.Container.js
orhantoy: Ok, it's fine to keep as is then.

orhantoy (approved):
I haven't reviewed in detail so I'm just approving to unblock you.

PR #144, opened 2023-01-12
fix(order_items): post order_items and shopping cart products amount in nav bar

RZhyvitskyi:
# Description:

# Jira:

### DIFF
--- a/packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
+++ b/packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
@@ -16,9 +16,14 @@ export const useShoppingCartContext = () => {
   return useContext(ShoppingCartContext);
 };
 
+const countAllProducts = (products) => {
+  return products.reduce((acc, product) => acc + product.quantity, 0);
+};
+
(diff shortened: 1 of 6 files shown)


patel-prakashkumar (dismissed):

packages/client/src/utils/ShoppingCartContext/ShoppingCartContext.jsx
```diff
@@ -16,9 +16,14 @@ export const useShoppingCartContext = () => {
   return useContext(ShoppingCartContext);
 };
 
+const countAllProducts = (products) => {
+  return products.reduce((acc, product) => acc + +product.quantity, 0);
```
orhantoy: `acc + +product.quantity` looks a bit weird. Do we need the extra plus?

RZhyvitskyi: Not realy now. Before this PR `product.quantity` was `typeof String`, that's why I made that. But not we don't need that.

orhantoy (approved):
Thanks for fixing this 🎉 💪

patel-prakashkumar (approved):

viedamayanti (approved):

PR #149, opened 2023-01-15
post-delete-end-point-of-favourites: 

27ManishaPatel:
post-delete-end-point-of-favourites

# Description:post & delete-end-point-of-favourites created with token 

# Jira:
https://hackyourfuture-dk.atlassian.net/jira/software/projects/CLASS22/boards/5?assignee=637964c39341d1f13604c560&selectedIssue=CLASS22-168

### DIFF
--- a/packages/server/api/controllers/favorites.controller.js
+++ b/packages/server/api/controllers/favorites.controller.js
@@ -54,12 +54,15 @@ const createFavorites = async (token, body) => {
 // delete
 const deleteFavorites = async (token, favoritesId) => {
   const userUid = token.split(' ')[1];
+
   const user = (await knex('users').where({ uid: userUid }))[0];
   if (!user) {
     throw new HttpError('User not found', 401);
   }
   try {
-    const deletedFav = knex('favorites').where({ id: favoritesId }).del();
+    const deletedFav = knex('favorites')
+      .where({ id: favoritesId, user_id: user.id })
+      .del();
     if (deletedFav === 0) {
       throw new HttpError('The favorites ID you provided does not exist.', 400);
     } else {
--- a/packages/server/api/controllers/favorites.controller.js
+++ b/packages/server/api/controllers/favorites.controller.js
@@ -32,11 +32,13 @@ const getFavoritesByUserId = async (token) => {
 };
 
 // post
-const createFavorites = async (body) => {
-  if (!body.user_id || !body.product_id) {
-    throw new HttpError('enter a value', 400);
-  }
+const createFavorites = async (token, body) => {
   try {
+    const userUid = token.split(' ')[1];
+    const user = (await knex('users').where({ uid: userUid }))[0];
+    if (!user) {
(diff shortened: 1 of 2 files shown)


packages/server/api/controllers/favorites.controller.js
```diff
@@ -32,11 +32,13 @@ const getFavoritesByUserId = async (token) => {
...
-  }
+const createFavorites = async (token, body) => {
   try {
+    const userUid = token.split(' ')[1];
+    const user = (await knex('users').where({ uid: userUid }))[0];
+    if (!user) {
+      throw new HttpError('User not found', 401);
+    }
     await knex('favorites').insert({
       user_id: body.user_id,
```
shpomp: I know this is from before, but I think this should come from the `user` and not the `body`, so `user.id `🙏🏽

shpomp (dismissed):
Please update all swagger so it can be tested out easily 🙏🏽

packages/server/api/controllers/favorites.controller.js
27ManishaPatel: Ohh yes right I will make a change

27ManishaPatel: Done ✅

packages/server/api/controllers/favorites.controller.js
```diff
@@ -50,9 +52,11 @@ const createFavorites = async (body) => {
...
-  if (!favoritesId) {
-    throw new HttpError('please enter ID to delete your favorite.', 400);
+const deleteFavorites = async (token, favoritesId) => {
+  const userUid = token.split(' ')[1];
+  const user = (await knex('users').where({ uid: userUid }))[0];
+  if (!user) {
+    throw new HttpError('User not found', 401);
   }
   try {
     const deletedFav = knex('favorites').where({ id: favoritesId }).del();
```
orhantoy: You should also check for the user ID here, otherwise any user can delete any favorite of another user.

packages/server/api/routes/favorites.router.js
```diff
@@ -95,8 +96,9 @@ router.post('/', (req, res, next) => {
  */
 
 router.delete('/:id', (req, res, next) => {
+  const { token } = req.headers;
   favoritesController
-    .deleteFavorites(req.params.id, req)
+    .deleteFavorites(token, req.params.id, req)
```
orhantoy: The `req` parameter is not used by the function, you only use the first two in the function, so I would just remove it.

packages/server/api/controllers/favorites.controller.js
27ManishaPatel: done ✅

packages/server/api/routes/favorites.router.js
27ManishaPatel: Done ✅

orhantoy (approved):

patel-prakashkumar (approved):

PR #150, opened 2023-01-15
fix(Modal): reset body overflow after closing

shpomp:
# Description:

Added 
`    document.body.style.overflow = 'visible';
` to Modal toggle function.

### DIFF
--- a/packages/client/src/components/Modal/Modal.component.js
+++ b/packages/client/src/components/Modal/Modal.component.js
@@ -9,6 +9,8 @@ const Modal = ({ toggle, open, title, children }) => {
     } else {
       document.body.style.overflow = '';
     }
+    // eslint-disable-next-line no-return-assign
+    return () => (document.body.style.overflow = '');
   }, [open]);
 
   if (!open) return null;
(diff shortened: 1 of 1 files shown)


orhantoy (commented):
I don't think this fix would work properly if the `open` prop is toggled. I was thinking more about something like this:

```js
const Modal = ({ toggle, open, title, children }) => {
  useEffect(() => {
    if (open) {
      document.body.style.overflow = 'hidden';
    } else {
      document.body.style.overflow = '';
    }
    return () => document.body.style.overflow = '';
  }, [open])
    
  if (!open) return null;
  // ...
```

packages/client/src/components/Modal/Modal.component.js
```diff
@@ -1,10 +1,17 @@
...
+import React, { useEffect } from 'react';
 import './modal.styles.css';
 
 const Modal = ({ toggle, open, title, children }) => {
+  useEffect(() => {
+    if (open) {
+      document.body.style.overflow = 'hidden';
+    } else {
+      document.body.style.overflow = '';
+    }
```
orhantoy, line 11: I would also include the cleanup function to handle the component being unmounted.

shpomp: Skipped it because linter is cursing 🙈

PR #151, opened 2023-01-15
Class22 167 choose dropdown options modal on click button buy now and add to cart

patel-prakashkumar:
# Description:
I have added modal on click buy Now and Add to Cart Button. 

# Jira:


shpomp (approved):
Nice consideration to check which ones are missing 🤓  👍🏽

PR #152, opened 2023-01-16
Fix: updating favorites should reflect in the UI

orhantoy:
# Description:

Currently the UI is not updated when the user creates or deletes a favorite. A full page refresh is needed. This PR tries to fix that by fetching the latest data when the user creates/deletes a favorite.


packages/client/src/containers/CollectionsPage/CollectionsPage.Container.js
```diff
@@ -16,29 +16,28 @@ export const CollectionsPage = () => {
...
+      headers: {
+        token: `token ${user?.uid}`,
+      },
+    })
+      .then((res) => res.json())
+      .then((items) => {
+        setProducts(items);
+        setIsLoading(false);
+      });
+  }, [selectedCategory, user]);
```
27ManishaPatel, line 36: well done 👍

27ManishaPatel (approved):

PR #153, opened 2023-01-16
feature(fix bug in Category endpoint) add user authantication

patel-prakashkumar:
# Description:
Refactor Category endpoint.
Add user authentication in category end point. 

# Jira:

### DIFF
--- a/packages/server/api/controllers/products.controller.js
+++ b/packages/server/api/controllers/products.controller.js
@@ -39,25 +39,23 @@ const getProductsByCategory = async (token, category, limit, offset) => {
     throw new HttpError('Category should be a String', 400);
   }
   try {
-    if (token) {
-      const userUid = token.split(' ')[1];
-      const user = (await knex('users').where({ uid: userUid }))[0];
-      products = await knex('products')
-        .select(
-          'products.id as id',
-          'products.name',
-          'products.description',
-          'products.price',
-          'categories.name as cname',
-          'favorites.id as favoritesID',
-        )
-        .leftJoin('categories', 'products.category_id', 'categories.id')
-        .offset(offset)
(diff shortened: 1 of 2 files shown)


packages/server/api/controllers/products.controller.js
```diff
@@ -33,22 +33,43 @@ const getProductById = async (id) => {
...
+          this.andOnVal('favorites.user_id', '=', `${user ? user.id : ''}`);
+        });
+    } else {
+      products = await knex('products')
+        .select(
+          'products.id as id',
+          'products.name',
+          'products.description',
+          'products.price',
+          'categories.name as cname',
```
orhantoy: This response will not include a `favoritesID` which I think will make it look like it is a favorite in the frontend. Could it work if you add `NULL as favoritesID` as an extra column here?

patel-prakashkumar: I have tried to add NULL as favoritesID as an extra column but Its not working.  Its working in sql query but not accepting null value in knex select query.

shpomp: My sleep-deprived understanding:
We never fall into else, as `token` is always true, because this is how we pass it:  
image

So, we always have a list where every product includes a `favoritesID`, event when not logged in:
image

Therefore, in `ProductList`, this checks out as false:
image

And we have no favourites in any category. 
Let's merge! :D

shpomp (dismissed):

packages/server/api/controllers/products.controller.js
orhantoy: I agree, let's remove the else part.

27ManishaPatel (dismissed):

PR #157, opened 2023-01-17
Fix(various): tweaks for better looks

shpomp:
Some small fixes making things look better. Never make a commit omelette like in this PR 🙏🏽 ❗ 
But this is an exception!

- Visual changes in confirmation and no-user modals
- "Back to Search" leads to "`/collections`"
- Call-to-action buttons in Home lead to "`/collections`"
- `<Button>` shadow on hover, ViewPageButton Unchanged
- "Add to cart" causes a smooth scroll-to-top
- some more (see commits)


packages/client/src/components/Body/Body.component.js
```diff
@@ -1,12 +1,13 @@
...
-import imgEarring from '../../assets/images/body-images/earring.png';
+import { Link } from 'react-router-dom';
 import imgCloset from '../../assets/images/body-images/closet.png';
+import imgEarring from '../../assets/images/body-images/earring.png';
+import imgGlass from '../../assets/images/body-images/glass.png';
+import imgHandbag from '../../assets/images/body-images/handbag.png';
+import imgMenAccessories from '../../assets/images/body-images/men-accessories.png';
+import imgWomenAccessories from '../../assets/images/body-images/women-accessories.png';
+import { Button } from '../Button/Button.component';
+import './Body.styles.css';
```
shpomp, line 10: These reorderings are my VS Code settings 🙈

Hailegebriel22 (dismissed):

